PR #5, opened 2016-09-05
Added docker

stephenwf:
To facilitate local testing when we decide on a search implementation.

@elrossco22

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM php:7.0-fpm
+
+RUN apt-get update && \
+    apt-get upgrade -y
+
+RUN apt-get install -y git curl wget zlibc zlib1g zlib1g-dev
+
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/bin/ --filename=composer
+#RUN curl -sS https://github.com/puli/cli/releases/download/1.0.0-beta10/puli.phar && \
+#    mv puli.phar /usr/local/bin/puli && \
+#    chmod 755  /usr/local/bin/puli
+
+
+RUN echo 'date.timezone = "Europe/London"' >> /usr/local/etc/php/conf.d/php.ini
+RUN docker-php-ext-install zip
+
+COPY composer.json /searchapi/composer.json
+COPY composer.lock /searchapi/composer.lock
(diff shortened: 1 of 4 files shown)


elrossco22:
👍

giorgiosironi:
I'm all for local testing, but we usually standardize on Vagrant machines. @lsh-0 what do you think?

lsh-0:
I dislike PHP for many reasons. One reason is that installing a specific version of PHP is a PITA. Another is having multiple applications with different PHP versions needing to co-exist together. Another is security. I'm in favour of Docker containers because it helps me manage the inherent shittiness of PHP but there is no support in builder for them yet. There have also been problems with race conditions with Ubuntu 14.04 + Docker that have not been fixed until 16.04. 

I'm of the opinion developers should be able to develop in whatever way is best for them, provided I can re-create what they're doing with Salt. If Docker is to be used for your testing then the version of PHP used will need to track the version we can install on a plain Ubuntu 14.04 AWS instance, which is PHP 7.0 from memory.

lsh-0:
I've been talking with @giorgiosironi and we already have a Salt formula for this. Any work you do with Docker will be duplicating what we have. I'm still in favour of you guys doing what works best for you, but please let us know if the environment changes significantly. We're adding you to the `maintainers.txt` that will alert you should our build break.

giorgiosironi:
I have added myself, @stephenwf and @elrossco22 to the `maintainers.txt`, which should notify us of any build break.

The `search--ci`, `search--end2end` and `search--prod` (this one not existing yet but identical to `search--end2end`) will be provisioned by https://github.com/elifesciences/search-formula which now you can also edit for minor changes. When there is software to install, like ElasticSearch, I will help getting the same setup there.

stephenwf:
Hi @lsh-0 @giorgiosironi 

Ah no this was not to replace the builder's environments, it was only to facilitate the upcoming search implementation and ensure while we are iterating during development other members can quickly spin up the correct versions on a per-branch basis whilst taking advantage of the speed of the recent native docker implementation on macOS.

I will just close this for now and maybe we can catch up when we're in Cambridge next week.

giorgiosironi:
:+1: for catching up in person

lsh-0:
I understood the intent entirely and I'm happy for you to work with Docker. Giorgio was concerned about duplication of efforts.

PR #6, opened 2016-09-05
Article responses

stephenwf:
@elrossco22 

That's all of the responses finished and tested. There is something strange about the time format of articles, also some of the articles may have too many fields, I will comb through the RAML tomorrow.

I've also hooked up the 'fake' query builder and search that will execute a search in a similar fashion to the api-mock (limited, but functional).


elrossco22:
👍

PR #11, opened 2016-09-07
Added gearman work and proof of concept test integration with api-sdk

stephenwf:
@elrossco22 The Gearman implementation

### DIFF
--- a/bin/do
+++ b/bin/do
@@ -27,7 +27,8 @@ function getBlogPost($page = 0, $perPage = 1)
     ];
 }
 
-function reverse($data) {
+function reverse($data)
+{
     return [
         'reverse',
         $data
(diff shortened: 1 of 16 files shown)


giorgiosironi:
Would you like me to add the two scripts as Upstart jobs, so that they are always running after a deploy in the `end2end` environment?

stephenwf:
I think this is going to change slightly once we have the symfony console component. Also the "do" command just now is just for testing the population from scratch (short and quick to type!). I think it will be renamed soon.

The ./bin/worker will spin up a single worker, so I think maybe some configuration will need to be plugged in to spin up X amount of workers to handle the jobs.

Also this requires Gearman server(s) which will also need to be put into configuration.

giorgiosironi:
I will:
- add a German job server to the instances
- set up the worker under upstart (our choice for long-running processes)
- add a configuration the number of workers (like for other projects, the code just needs to worry about providing one worker script, the formula can launch it X times and we already have the code to do that)

stephenwf:
That sounds good, do you want to stick with ./bin/worker to spin up a worker?

giorgiosironi:
Since this code is not in develop yet, I will use `sleep 60` as a dummy command, and then you can configure it to anything once the code is stable.

elrossco22:
👍

PR #13, opened 2016-09-15
Added Simple API workflow structure

stephenwf:
@elrossco22


stephenwf:
@thewilkybarkid This request the fix we were talking about this morning with the JsonEncoder, do you want me to raise a PR with the ApiSdk repo?

tests/src/Search/Workflow/GetSerializer.php
```diff
@@ -0,0 +1,43 @@
...
+
+trait GetSerializer
+{
+    private $serializer;
+
+    public function getSerializer()
+    {
+        if (!method_exists($this, 'getHttpClient')) {
+            throw new LogicException('You need to include the HttpClient to initialize the serializer');
+        }
```
thewilkybarkid: This trait should `use` the other one.

thewilkybarkid:
@stephenwf Please do

tests/src/Search/Workflow/GetSerializer.php
stephenwf: I had that initially but then when I was creating the Mock trait here: 
https://github.com/elifesciences/search/pull/13/files/20e8dc5dd227b86889fc3e2c968b6c0ad3146efe#diff-92d16925d99a8a21d6f7bb0638029052R42

I realised I had a double-dependency, what would you suggest?

src/Search/Api/Elasticsearch/ElasticSearchResponse.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+namespace eLife\Search\Api\Elasticsearch;
+
+
+use eLife\Search\Api\Query\int;
```
elrossco22: Can we remove this, if it isn't used please

PR #14, opened 2016-09-27
Added CLI class workflow

stephenwf:
@elrossco22


giorgiosironi:
For future reference, what is the difference between `local` and `dev`?

stephenwf:
Ah local was a configuration that I ignored, I can swap that for dev if it makes more sense?

giorgiosironi:
`dev` is whatever will be used inside the Vagrant machine provided by builder. I guess you're using `local` for some other setup? But if you want to overwrite the `dev` files instead with that we will have a single configuration to use on our laptops (as opposed to `ci`, `end2end`, `prod` which are used on EC2 instances only).

PR #15, opened 2016-09-27
Added logger for workflow (to allow cli logging)

stephenwf:
@elrossco22


tests/src/Search/Workflow/BlackholeLogger.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace tests\eLife\Search\Workflow;
+
+use Psr\Log\LoggerInterface;
+
+trait BlackholeLogger
+{
+    public function getLogger()
+    {
+        return new class() implements LoggerInterface {
```
thewilkybarkid: PSR-3 comes with a `NullLogger`.

stephenwf: One day ill find a use for an anonymous class!

PR #18, opened 2016-10-10
CLI Commands

stephenwf:
@elrossco22 

@giorgiosironi I've added what will become the final worker command. Currently its run like this:
```bash
$ php bin/console gearman:worker
```

but once we have more pieces in place (the different models)

```bash
$ php bin/console gearman:worker --workflows=blog-articles,interviews,articles
```

And these spin up one single worker. The Gearman worker is tied to this process and when the kill signal is sent it will not leave child processes.

The remaining challenge is monitoring memory and possibly a timeout and for the worker to tear itself down. I've done some testing (you can see the spawn command), but there are orphan processes still remaining. I've explored a few different ways of self-destructing the workers, but I want to take a step back for now and come back with fresh pair of eyes.


stephenwf:
In addition to the `gearman:worker` command I've added `debug:params` which will return the configuration of the selected environment, e.g.
```
$ php bin/console debug:params --env=prod
```

and also a clear cache
```
$ php bin/console cache:clear
```

which does not currently respect the environments and nukes the whole thing (a todo).

giorgiosironi:
I will change the configuration in `search-formula` to run this command.

PR #19, opened 2016-10-11
Added more workflows

stephenwf:
@elrossco22


config/extra/gearman-shim.php
```diff
@@ -0,0 +1,23 @@
+<?php
```
elrossco22: Should we perhaps rename this file so it reflects the class name?

stephenwf:
Also contains some onboarding instructions.

Before:
screen shot 2016-10-11 at 14 20 50

After:
screen shot 2016-10-11 at 14 20 55

stephenwf:
Also for local environment that may not have Gearman PECL client we now show a warning (can be suppressed in configuration)

screen shot 2016-10-11 at 14 28 27

PR #21, opened 2016-10-12
Adding an unused argument to distinguish processes

giorgiosironi:
This is a common practice on our long-running process, to quickly see if
there are duplicates or some of the processes have crashes e.g. when
running `ps faxww` you'll see this parameter in the command lines used
to start the workers.


stephenwf:
Cool, so we shouldn't need to use that ID, just have it as part of the command?

giorgiosironi:
No need to use it, although if you want to include it in the logs during startup, along with the process id, it's a nice information to have when debugging.

PR #22, opened 2016-10-12
Elasticsearch

stephenwf:
@elrossco22


stephenwf:
So far (providing you have everything set up and configured) you can import Events and Interviews from the API (api-dummy) and they will appear in elastic search under /search which you can perform the RAML-defined queries against. (I've only tested `for={}` just now).

PR #26, opened 2016-11-02
Elasticsearch Mappings [WIP]

stephenwf:
@elrossco22  This is the start of a chunk of work to optimise, set weightings and identify polymorphic fields (and therefore incompatible with ES).

Work in progress, but feedback appreciated.

### DIFF
--- a/src/Search/Api/Elasticsearch/Command/BuildIndexCommand.php
+++ b/src/Search/Api/Elasticsearch/Command/BuildIndexCommand.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace eLife\Search\Api\Elasticsearch\Command;
+
+use Symfony\Component\Console\Command\Command;
+
+class BuildIndexCommand extends Command
+{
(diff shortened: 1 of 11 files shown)


stephenwf:
@elrossco22 

I've updated the mappings here with an import feature that will drop the elasticsearch index and create a new one based on configuration. This will probably only be used for provisioning and development. 

Hot-deployments of new schemas would most likely have to have some sort of other deployment process, where a second index is created along side, and then swapped out. I'm working under the assumption that this sort of hot-deployments are out of scope for now. (@giorgiosironi  can you confirm).

giorgiosironi:
Don't know the exact requirements, but we definitely should postpone any hot deployment to after we have this running in production.

stephenwf:
@giorgiosironi I've merged this now as the updated index creation is needed for the saltstack.

I will continue to improve the process of initial set up of the database in subsequent PRs.

PR #27, opened 2016-11-03
Fixed RAML bug on E2E environment

stephenwf:
@giorgiosironi @elrossco22

### DIFF
--- a/src/Search/Api/Query/RamlRequirement.php
+++ b/src/Search/Api/Query/RamlRequirement.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace eLife\Search\Api\Query;
+
+trait RamlRequirement
+{
+    public static $root = __DIR__.'/../../../../vendor/elife/api/dist/samples/';
+
+    public function getFixture(string $name) : string
+    {
+        $file = self::$root.$name;
+
+        return file_exists($file) ? file_get_contents($file) : null;
+    }
+}
--- a/src/Search/Api/Query/MockQueryBuilder.php
+++ b/src/Search/Api/Query/MockQueryBuilder.php
@@ -2,8 +2,6 @@
 
 namespace eLife\Search\Api\Query;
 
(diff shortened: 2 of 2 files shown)


giorgiosironi:
Shouldn't the file also be removed from `test/`? Unless the two versions of `RamlRequirement` are different.

stephenwf:
@giorgiosironi Yeh so the one in tests will mark a test as incomplete if the YAML doesn't exist, so I just copied it across and removed the test-specific stuff. This will be removed quite soon once the ES instance is available

giorgiosironi:
:+1:

src/Search/Api/Query/RamlRequirement.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace eLife\Search\Api\Query;
+
+trait RamlRequirement
+{
+    public static $root = __DIR__.'/../../../../vendor/elife/api/dist/samples/';
```
thewilkybarkid: Hmm, we should use Puli rather than hardcode such a path. But I don't think anything should rely on those samples anyway.

stephenwf: At the moment the unit tests use them, and happy to replace with puli, but not exactly sure how to go about that. (I remember asking about the puli paths for the raml)

As for this part of the site, yes its only for the interim while we are getting the ES instance set up.

thewilkybarkid: The relevant Puli mapping is https://github.com/elifesciences/api-raml/blob/develop/puli.json#L4-L7. But in this case I think it's a false abstraction, and the fixtures should live here (sounds like the test could start to fail if a legitimate change is made in the other repo).

stephenwf: Yeh that was the aim was for breaking changes to cause these particular tests to fail. It should hopefully help with syncing prior to version 1.0.

PR #28, opened 2016-11-03
[Review] Search Queue initial interfaces for review

stephenwf:
@elrossco22 @giorgiosironi 

Just wanted to make sure logic was sound on the flow of this before creating the static example implementation.

### DIFF
--- a/src/Search/Queue/WatchableQueue.php
+++ b/src/Search/Queue/WatchableQueue.php
@@ -26,7 +26,7 @@ interface WatchableQueue
      * Mock: Remove item completely.
      * SQS: this will delete the item from the queue.
      */
-    public function commit(QueueItem $item) : bool;
+    public function commit(QueueItem $item);
 
     /**
      * This will happen when an error happens, we release the item back into the queue.
--- a/src/Search/Queue/WatchableQueue.php
+++ b/src/Search/Queue/WatchableQueue.php
@@ -18,7 +18,7 @@ interface WatchableQueue
      * Mock: Move to separate "in progress" queue.
      * SQS: this will change the timeout of the in-memory item.
      */
-    public function dequeue(int $timeout = 30) : QueueItem;
+    public function dequeue(int $timeoutOverride = null) : QueueItem;
 
     /**
      * Commits to removing item from queue, marks item as done and processed.
--- a/src/Search/Queue/WatchableQueue.php
+++ b/src/Search/Queue/WatchableQueue.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace eLife\Search\Queue;
+
+interface WatchableQueue
+{
+    /**
+     * Adds item to the queue.
+     *
+     * Mock: Add item to queue.
+     * SQS: This will set the queue item into the memory slot for re-processing.
+     */
+    public function enqueue(QueueItem $item) : bool;
+
+    /**
+     * Starts process of removing item.
+     *
+     * Mock: Move to separate "in progress" queue.
+     * SQS: this will change the timeout of the in-memory item.
+     */
+    public function dequeue(int $timeout = 30) : QueueItem;
+
+    /**
+     * Commits to removing item from queue, marks item as done and processed.
+     *
+     * Mock: Remove item completely.
+     * SQS: this will delete the item from the queue.
+     */
+    public function commit(QueueItem $item) : bool;
+
+    /**
+     * This will happen when an error happens, we release the item back into the queue.
+     *
+     * Mock: re-add to queue.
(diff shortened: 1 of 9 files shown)


elrossco22:
Looks good for a WIP

src/Search/Queue/WatchableQueue.php
```diff
@@ -0,0 +1,46 @@
...
+     * SQS: This will set the queue item into the memory slot for re-processing.
+     */
+    public function enqueue(QueueItem $item) : bool;
+
+    /**
+     * Starts process of removing item.
+     *
+     * Mock: Move to separate "in progress" queue.
+     * SQS: this will change the timeout of the in-memory item.
+     */
```
giorgiosironi: I guess the `timeout` is the visibility timeout, makes sense to be able to override it if there is a particularly heavy message to process, but given the queue is just one this could also go into the configuration of the `WatchableQueue` object to make the interface simpler.

src/Search/Queue/WatchableQueue.php
```diff
@@ -0,0 +1,46 @@
...
+    public function dequeue(int $timeout = 30) : QueueItem;
+
+    /**
+     * Commits to removing item from queue, marks item as done and processed.
+     *
+     * Mock: Remove item completely.
+     * SQS: this will delete the item from the queue.
+     */
```
giorgiosironi: mandatory for SQS (and hence in the interface), happy paths are a sequence of `dequeue()` followed by some work and `commit()`

src/Search/Queue/WatchableQueue.php
```diff
@@ -0,0 +1,46 @@
...
+namespace eLife\Search\Queue;
+
+interface WatchableQueue
+{
+    /**
+     * Adds item to the queue.
+     *
+     * Mock: Add item to queue.
+     * SQS: This will set the queue item into the memory slot for re-processing.
+     */
```
stephenwf: Yes absolutely, just needed something for testing!

stephenwf:
@giorgiosironi @elrossco22  I've just added a mock implementation (rough, but it is working locally)

src/Search/Queue/WatchableQueue.php
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * Mock: Remove item completely.
+     * SQS: this will delete the item from the queue.
+     */
+    public function commit(QueueItem $item) : bool;
+
+    /**
+     * This will happen when an error happens, we release the item back into the queue.
+     *
+     * Mock: re-add to queue.
```
giorgiosironi: I'm not sure what the memory slot is in the case of SQS. When a message is returned by the `receiveMessage` operation, it can be returned by `dequeue()` without changing the state of the `WatchableQueue` object. A "memory slot" may act as a buffer but what is the purpose?

src/Search/Queue/WatchableQueue.php
giorgiosironi: There is also the option of committing the message as deleted from the queue after the job is added to Gearman, rather than waiting for the whole job to be executed. That would free you from having to notify SQS (not sure how too) when the Gearman job is finished; it would only work reliably if we persist Gearman jobs into a database (IIRC MySQL can be configured) rather than keeping them in memory, to avoid data loss in case of crash.

About the process being terminated, the default behavior of a PHP process is to stop immediately; what happens is there will be no `commit()` call happening for the dequeued item, and after its visibility timeout expires it will be accessible to the next process taking items off the queue. This is at-least-once delivery and it's safe. It can be optimized by intercepting the SIGTERM signal and setting the visibility timeout to 0 when you receive it, but PHP works in a single thread so that would mean putting the `pcntl_signal_dispatch()` call that triggers (intercepted) signals delivery somewhere in the loop... it gets complex quickly.

src/Search/Queue/WatchableQueue.php
giorgiosironi: I would suggest going back to the queue, if only to keep it simple :) to improve throughput we can always add worker processes

src/Search/Queue/WatchableQueue.php
stephenwf: I do think that hooking Gearman up to MySQL would simplify this quite a lot, make it safer anyway. That would allow Gearman to handle its own recovery and re-trying, like it does currently with the batch processing.

Is the MySQL integration with Gearman transparent in terms of the implementation, or would we have to change some specifics?

giorgiosironi: Looks like it's configuration for the daemon, the API stays the same. We would use RDS in production/end2end.

PR #30, opened 2016-11-04
SQS Queue Implementation

stephenwf:
@elrossco22 @giorgiosironi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-.graniterc
 /bin/commands
 /build/
 /vendor/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.graniterc
+/bin/commands
 /build/
 /vendor/
 /cache/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/bin/commands
 /build/
 /vendor/
 /cache/
--- a/config/local.example.php
+++ b/config/local.example.php
@@ -12,4 +12,11 @@ return [
     'ttl' => 0,
     'gearman_servers' => ['elife_gearman_1'],
     'gearman_auto_restart' => true,
+    'aws' => [
+        'mock_queue' => false,
+        'queue_name' => 'eLife-search',
+        'key' => '-----------------------',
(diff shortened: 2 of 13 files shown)


stephenwf:
Work in progress due to stubbed queue item still being used for test data. To be removed.

.gitignore
```diff
@@ -1,3 +1,5 @@
+.graniterc
```
giorgiosironi: what's this?

src/Search/Gearman/Command/QueueCommand.php
```diff
@@ -120,23 +125,29 @@ public function checkStatus(InputInterface $input, LoggerInterface $logger)
 
     public function loop(InputInterface $input, LoggerInterface $logger)
     {
+        $logger->info('Loop start... [');
         $topic = $input->getArgument('topic');
         $timeout = $input->getOption('queue-timeout');
-        $logger->info('Hello queue. topic: '.$topic);
+        $logger->info('-> Hello queue. topic: '.$topic);
```
giorgiosironi: For parameterizable logging in general, I suggest to use the $context argument https://github.com/Seldaek/monolog/blob/master/doc/01-usage.md#using-the-logging-context which would make this easier to search from log aggregators

.gitignore
stephenwf: Ah sorry I'll remove that

config/local.example.php
```diff
@@ -12,4 +12,11 @@
     'ttl' => 0,
     'gearman_servers' => ['elife_gearman_1'],
     'gearman_auto_restart' => true,
+    'aws' => [
+        'mock_queue' => false,
```
stephenwf: okay that would work, maybe it would be possible to set up a hybrid system where either the configuration can be set up in the file (maybe for local, I'm using a specific account for testing), or in the `.aws/credentials` file.

src/Search/Gearman/Command/QueueCommand.php
stephenwf: Okay changed this file to use that, I imagine when running verbose it will show the details

PR #35, opened 2016-11-09
Removed old index creation

stephenwf:
@giorgiosironi 

This should be a successful build using the new configuration.


giorgiosironi:
Two things:
1) PRs build only on `search--ci` but not on `search--end2end` as the pipeline stops short
2) I need to update the secret `builder-private` repository to add the new .sls files to the list to trigger them.

giorgiosironi:
Rerunning the build now that the new states are configured.

giorgiosironi:
Failure is:
`09:58:45 [54.210.183.237] out:                   - E: Version '5.0' for 'elasticsearch' was not found`
You may want to add a `refresh: True` to the `pkg.installed` state, which would force it to run an `apt-get update` (if it's not already there). Also check the precise name of the version, there may some suffix to `5.0` that we don't know.

stephenwf:
Just opened up a new branch now and running the stack again to find which version is installed. Ill add the refresh parameter

stephenwf:
@giorgiosironi Looks like its now the memory issue again for elasticsearch, does the CI use AWS for the boxes?

giorgiosironi:
Yes we are using t2.small instances which have 2GB of RAM. Should we go to 4GB?

giorgiosironi:
Answer: yes

giorgiosironi:
`search--ci` switched to 4GB of RAM, retrying build.

stephenwf:
Success 🍾

giorgiosironi:
Works on `search--ci`, but let me recreate `search--end2end` before merging.

giorgiosironi:
Also `search--end2end` has been rebuilt now.

PR #36, opened 2016-11-11
Added more workflow tests

stephenwf:
@elrossco22 

Hi @thewilkybarkid, I'm having some issues with checking the equality of properties on different objects. Specifically those with nested promises. I've created a async wrapper around the PHPUnit assertions that will await on values that are passed in for assertions, but this does not unwrap in a cascading manner. I have a few questions:

1) There has not been any network requests in the tests, simply `Object -> JSON -> Object`, when the object is denormalised, is it possible to not wrap concrete values in promises?

2) Is there a way to unwrap the whole object recursively?

### DIFF
--- a/src/Search/Api/Response/SearchResponse.php
+++ b/src/Search/Api/Response/SearchResponse.php
@@ -49,8 +49,10 @@ final class SearchResponse implements HasHeaders
         $this->items = $items;
         // @todo remove this hack!
         $this->items = array_map(function ($item) {
-            if ($item->image) {
-                $item->image = $item->image->https();
+            if (property_exists($item, 'image')) {
+                if ($item->image) {
+                    $item->image = $item->image->https();
+                }
             }
             $item->statusDate = new DateTime();
 
--- a/tests/bootstrap
+++ b/tests/bootstrap
@@ -12,7 +12,9 @@ namespace {
     require_once __DIR__ . '/../vendor/elife/api-sdk/test/PuliAwareTestCase.php';
     require_once __DIR__ . '/../vendor/elife/api-sdk/test/ApiTestCase.php';
     require_once __DIR__ . '/../vendor/elife/api-sdk/test/ApiSdkTest.php';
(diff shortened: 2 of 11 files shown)


thewilkybarkid:
Are these all `Sequence`s? If so, try `->toArray()` and compare them.

stephenwf:
So the sequences themselves contain other promises, like images, body content (more sequences).`->toArray()` does not appear to cascade in sequences at the moment

stephenwf:
The tests themselves are testing the fairly simple `object == deserialize(serialize(object))`

thewilkybarkid:
Without having dived into what's going on, should you be testing that? The SDK provides that for you, so these no need to test it here? /cc @giorgiosironi

giorgiosironi:
I'll take a look at the code to see the problem, also because there are some cycles between entities (a Collection containing a Podcast pointing to another Collection) that could a full serialization an infinite loop if used in production.

stephenwf:
The main purpose was for potential future breaking changes in methods on these objects we use (indirectly for constructing search results) cause tests to fail

thewilkybarkid:
The SDK has to do something that sounds the same: https://github.com/elifesciences/api-sdk-php/blob/master/test/TestCase.php#L15-L37

giorgiosironi:
I understand these are contract tests on the `serialize()`/`deserialize()` pair. I wouldn't suggest to reimplement or borrow the comparison code here, as that is tested in the SDK itself. Wanting to place a basic check on serialization as a feature, you could stick to:
```
        $serialized = $this->workflow->serialize($collection);
        $deserialized = $this->workflow->deserialize($serialized);
        $this->assertInstanceOf(get_class($collection), $deserialized);
        $final_serialized = $this->workflow->serialize($deserialized);
        $this->assertJsonStringEqualsJsonString($serialized, $final_serialized);
```
if that does not require to mock too many HTTP calls (best case is no HTTP calls).

stephenwf:
Okay that makes sense, ill change to use that for these tests, thanks

PR #39, opened 2016-11-15
Fixed configuration again

stephenwf:
@giorgiosironi


giorgiosironi:
:+1: region is not sensitive so we can keep it here for now

PR #41, opened 2016-11-16
Added more api urls

stephenwf:
@giorgiosironi @elrossco22


giorgiosironi:
Hopwefully they will become https later, but :+1: now

giorgiosironi:
Umask is only part of the solution, see previous comments.

PR #44, opened 2016-11-30
Logging stack traces on failed imports

giorgiosironi:
This should really use Monolog, so that we can send the logs also to files and format them appropriately, etc.

This is the minimal change to be able to debug what is happening in the end2end environment.

Unfortunately there seem to be no generic way of dumping the identifier for a Model object from the SDK, so I can't put a `getId()` in the message as some Models do not have it.

### DIFF
--- a/src/Search/Gearman/Command/ApiSdkCommand.php
+++ b/src/Search/Gearman/Command/ApiSdkCommand.php
@@ -133,7 +133,7 @@ final class ApiSdkCommand extends Command
                     $this->task($task, $normalized);
                 }
             } catch (Throwable $e) {
-                $logger->alert("Error on a ".get_class($item), ['exception' => $e]);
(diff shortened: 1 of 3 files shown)


stephenwf:
Maybe we could add a __toString() to the models in Api SDK (ping @thewilkybarkid)

thewilkybarkid:
Similar to https://github.com/elifesciences/api-sdk-php/pull/76, maybe `Model` should define a `getIdentifier()` (which is separate to`getId()`, `getNumber()` or whatever).

(This _could_ be a value object of `type` (eg 'podcast-episode', 'research-article') + identifier.)

giorgiosironi:
Ok, agreed. In the mean time the stack trace will be useful and I will insert some debug commands manually when needed.

giorgiosironi:
Debugging the build before review.

giorgiosironi:
Ready now.

PR #45, opened 2016-11-30
Upgrade api-sdk to get UnknownReference support

giorgiosironi:
Empirically this work to solve errors during `gearman:import all`. Let's see if the test suite is green.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "description": "Search API gateway",
     "require": {
         "elife/api-validator": "dev-master#6dc70d7",
-        "elife/api-sdk": "dev-master#b2f591b844",
+        "elife/api-sdk": "dev-master#ea6d0d1",
         "elife/api": "dev-master#b3b1321",
         "silex/silex": "^2.0",
         "jms/serializer": "^1.3",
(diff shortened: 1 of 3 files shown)


giorgiosironi:
Lots of `Error: Class 'eLife\ApiSdk\Serializer\Block\FileNormalizer' not found`, what else should we upgrade?

stephenwf:
Ah this will probably need to be updated with the new normalizer:
https://github.com/elifesciences/search/blob/develop/tests/src/Search/Workflow/GetSerializer.php

giorgiosironi:
That should be provided by the SDK right? Even if this is customized, I can add a method there to remove the duplication.

stephenwf:
It's a bit of a pain, in the API SDK its in here: https://github.com/elifesciences/api-sdk-php/blob/master/src/ApiSdk.php

We could instantiate the class and try and pull the serializer out and set it on our object, I just didn't feel this was the cleanest solution.

giorgiosironi:
Making a comparison of the two lists, I see that some Normalizers have changed location (`FileNormalizer` moved outside of `Block` and more are missing besides it, like the `UnknownReferenceNormalizer`. From `ApiSdk` it's not easy to extract the `Serializer` without being very invasive (extracting a trait), I'll try to pull it out from there with the goal of having it always working right after the `composer update` on `elife/api-sdk`.

giorgiosironi:
Ok, works now.

stephenwf:
Looks good, much cleaner 👍

PR #46, opened 2016-11-30
Umask must make files group-writable

giorgiosironi:
So that www-data and elife can delete and rewrite files from one another. This will also require all files to be created in the www-data group


giorgiosironi:
Goes hand-in-hand with https://github.com/elifesciences/search-formula/commit/e7d84e4b51b96ca7da606dd6225008d9c8bac960

PR #47, opened 2016-11-30
Importing everything on search--end2end

giorgiosironi:
Also checking that all jobs in end2end are executed


giorgiosironi:
Let's try it, most of it is not executed on `ci` apart from the waiting script.

PR #48, opened 2016-12-07
ELPP-1763 Logging, Bugs and Article type

stephenwf:
@giorgiosironi  @elrossco22

### DIFF
--- a/src/Search/Api/SearchResultDiscriminator.php
+++ b/src/Search/Api/SearchResultDiscriminator.php
@@ -12,7 +12,6 @@ use JMS\Serializer\EventDispatcher\PreSerializeEvent;
 final class SearchResultDiscriminator implements EventSubscriberInterface
 {
     public static $articleTypes = [
-        'article',
         'correction',
         'editorial',
         'feature',
--- a/src/Search/Api/SearchResultDiscriminator.php
+++ b/src/Search/Api/SearchResultDiscriminator.php
@@ -12,6 +12,7 @@ use JMS\Serializer\EventDispatcher\PreSerializeEvent;
 final class SearchResultDiscriminator implements EventSubscriberInterface
 {
     public static $articleTypes = [
+        'article',
         'correction',
         'editorial',
         'feature',
--- a/src/Search/Workflow/BlogArticleWorkflow.php
+++ b/src/Search/Workflow/BlogArticleWorkflow.php
@@ -111,8 +111,10 @@ final class BlogArticleWorkflow implements Workflow
             // That blog article is valid JSON.
             $this->validator->validateSearchResult($result, true);
         } catch (Throwable $e) {
-            $this->logger->alert($e->getMessage());
-            $this->logger->alert('BlogArticle<'.$id.'> rolling back');
+            $this->logger->alert('BlogArticle<'.$id.'> rolling back', [
+                'message' => $e->getMessage(),
+                'exception' => $e,
+            ]);
             $this->client->deleteDocument($type, $id);
             // We failed.
             return self::WORKFLOW_FAILURE;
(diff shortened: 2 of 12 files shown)


src/Search/Workflow/BlogArticleWorkflow.php
```diff
@@ -111,8 +111,10 @@ public function postValidate(string $type, string $id) : int
             // That blog article is valid JSON.
             $this->validator->validateSearchResult($result, true);
         } catch (Throwable $e) {
-            $this->logger->alert($e->getMessage());
-            $this->logger->alert('BlogArticle<'.$id.'> rolling back');
+            $this->logger->alert('BlogArticle<'.$id.'> rolling back', [
+                'message' => $e->getMessage(),
+                'exception' => $e,
```
giorgiosironi: good to have exceptions in the context wherever they are caught

src/Search/Api/SearchResultDiscriminator.php
```diff
@@ -12,6 +12,7 @@
 final class SearchResultDiscriminator implements EventSubscriberInterface
 {
     public static $articleTypes = [
+        'article',
```
stephenwf: Okay so just so I'm clear:

SQS sends `{type: article}` and the real data is `{type: research-article}` or any of the others in the list.

giorgiosironi:
Yes, the SQS `type` varies as `article`, `event`, ...: https://github.com/elifesciences/builder/blob/master/projects/elife.yaml#L497
while the `/articles/...` API `type` varies as `correction`, `research-article`, ...: https://github.com/elifesciences/api-raml/blob/master/src/snippets/article.v1.yaml#L12

giorgiosironi:
Merge at will :+1: 
The end2end environment may be unstable today, but I will check that this gets through there so we can try to start the queue:watch.

giorgiosironi:
So this got to end2end, where the smoke first failed, then hang:
```
elife@end2end--search:~$ curl -v http://localhost/search
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /search HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost
> Accept: */*
> 
^C
```
Investigating.

giorgiosironi:
`curl -v http://localhost/foo` returns 404, so it's something in the code like an external request or a connection to a database.

stephenwf:
It looks like the only code that could cause that effect is the call to Elasticsearch.

stephenwf:
Could it have still been running at that point?

giorgiosironi:
```
{"error":"cURL error 7: Failed to connect to end2end--gateway.elifesciences.org port 80: Connection refused (see http:\/\/curl.haxx.se\/libcurl\/c\/libcurl-errors.html)",
"trace":"#0 \/srv\/search\/vendor\/guzzlehttp\/promises\/src\/Promise.php(203): eLife\\ApiClient\\HttpClient\\Guzzle6HttpClient-\u003EeLife\\ApiClient\\HttpClient\\{closu
re}(Object(GuzzleHttp\\Exception\\ConnectException))\n#1 
```
so it's connecting to the gateway, which as far as I can tell is needed to populate responses. Is there a call in the smoke tests that doesn't get outside of this EC2 instance? It's needed to validate that the deploy worked on the current node rather than failing because of external dependencies.

giorgiosironi:
I would say that an empty result like `curl -v localhost/search?for=inexistentinexistentinexistent` should not depend on the gateway to work?

giorgiosironi:
We can always introduce a `/ping` route for this use case (just returns 200).

stephenwf:
Addded here: https://github.com/elifesciences/search/pull/49

PR #49, opened 2016-12-07
Ping

stephenwf:
@giorgiosironi


giorgiosironi:
:+1:

giorgiosironi:
Ok, now it only connects to the gateway during the import and not during the smoke tests. I will make sure the `end2end` environment is brought up correctly for this build.

PR #50, opened 2016-12-12
Stable smoke test

giorgiosironi:
This uses api-gateway and journal-cms behind it (at least), but should be pretty stable:
- in `ci`, with api-dummy
- in `end2end`, with the other end2end machines (just tried it)
- in `prod`, with the other prod machines


giorgiosironi:
This currently breaks on end2end due to the `cache/` folder problems, so better to break it directly in the build of search than to find out when other projects reach that environment.

PR #52, opened 2016-12-13
Added basic fixtures for current tests

stephenwf:
When developing tests simply pass "true" to `getHttpClient()` and it will generate the fixtures if you have a local copy of the dummy/data source available. This should be removed before committing. 

@elrossco22


stephenwf:
Closing for now.

PR #55, opened 2016-12-13
ELPP-1929 Added monolog 

stephenwf:
@giorgiosironi There is a configuration option now for loggly. Let me know whats the best way for us to get this key into the configuration.

@elrossco22

### DIFF
--- a/config/local.example.php
+++ b/config/local.example.php
@@ -10,6 +10,8 @@ return [
     'elastic_servers' => ['http://elife_search_elasticsearch:9200'],
     'annotation_cache' => false,
     'ttl' => 0,
+    'file_log_path' => __DIR__.'../var/logs/debug.log',
+    'file_error_log_path' => __DIR__.'../var/logs/error.log',
     'gearman_servers' => ['elife_gearman_1'],
     'gearman_auto_restart' => true,
     'aws' => [
--- a/config/local.example.php
+++ b/config/local.example.php
@@ -10,7 +10,7 @@ return [
     'elastic_servers' => ['http://elife_search_elasticsearch:9200'],
     'annotation_cache' => false,
     'ttl' => 0,
-    'file_log_path' => __DIR__.'../var/logs/debug.log',
+    'file_log_path' => __DIR__.'../var/logs/all.log',
     'file_error_log_path' => __DIR__.'../var/logs/error.log',
     'gearman_servers' => ['elife_gearman_1'],
     'gearman_auto_restart' => true,
--- a/src/Search/Kernel.php
+++ b/src/Search/Kernel.php
@@ -68,8 +68,6 @@ final class Kernel implements MinimalKernel
 
     public function __construct($config = [])
     {
-        // perms.
-        umask(002);
         $app = new Application();
         // Load config
         $app['config'] = array_merge([
--- a/src/Search/Kernel.php
+++ b/src/Search/Kernel.php
@@ -81,8 +81,8 @@ final class Kernel implements MinimalKernel
             'elastic_servers' => ['http://localhost:9200'],
             'elastic_index' => 'elife_search',
             'gearman_auto_restart' => true,
-            'file_log_path' => self::ROOT.'/var/logs/debug.log',
-            'file_error_log_path' => self::ROOT.'/var/logs/error.log',
+            'file_log_path' => null,
+            'file_error_log_path' => null,
             'loggly_key' => null,
             'aws' => [
                 'credential_file' => false,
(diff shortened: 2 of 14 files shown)


stephenwf:
@giorgiosironi  I had to remove the filesystem log, I will update the formula and move the logs and cache into that var folder now that we have more than one place to write

giorgiosironi:
The only configuration for Loggly accessible to project formulas is whether it is enabled:
https://github.com/elifesciences/journal-formula/blob/master/salt/journal/config/etc-syslog-ng-conf.d-journal.conf#L27
We use syslog-ng, a daemon, to tail local files and send them to Loggly. This ensures if Loggly is down the application does not depend on it. What you need to put here should clone the journal configuration, which means adding some configuration for syslog-ng that will take the files as source and link them with the already existing (should be there because of builder-base-formula) `destination(d_loggly)`.

stephenwf:
@giorgiosironi So we could set up a normal log file and set up the rest in the salt stack? That makes it easier on this side, it doesn't need to know about loggly.

giorgiosironi:
Yes, the responsibility of the PHP code should just be to write to a single file (more than one file if you find it useful) in append mode, so that all converge there. However, this file should use the JSON formatter (again, follow what journal does) so that we stream each line of this file as a new log entry on Loggly and it gets parsed and filterable on the fields in the context.

src/Search/Kernel.php
```diff
@@ -178,13 +179,14 @@ public function dependencies(Application $app)
...
+                $stream->setFormatter(new LineFormatter());
+                $logger->pushHandler($stream);
             }
             if ($app['config']['file_error_log_path']) {
-                $logger->pushHandler(new StreamHandler($app['config']['file_error_log_path'], Logger::ERROR));
-            }
-            if ($app['config']['loggly_key']) {
-                $logger->pushHandler(new LogglyHandler($app['config']['loggly_key']));
+                $stream = new StreamHandler($app['config']['file_error_log_path'], Logger::ERROR);
+                $stream->setFormatter(new LineFormatter());
```
stephenwf: Okay I've switched to the json formatter, I couldn't see that configuration in the journal configuration, but maybe it was just the dev logs I was looking at.

giorgiosironi:
:+1: merge when green and we'll see what is written in the testing environments and production

config/local.example.php
```diff
@@ -10,6 +10,8 @@
     'elastic_servers' => ['http://elife_search_elasticsearch:9200'],
     'annotation_cache' => false,
     'ttl' => 0,
+    'file_log_path' => __DIR__.'../var/logs/debug.log',
```
giorgiosironi: if it's called `debug.log` it should have `debug` statements in it, not just `info`?

giorgiosironi: otherwise, name it `info.log` or `all.log` to differentiate it from the other, more selective one?

stephenwf:
@giorgiosironi I'm getting: 
```
Fatal error: Uncaught JMS\Serializer\Exception\InvalidArgumentException: The cache directory "/srv/search/src/Search/../../var/cache" is not writable.
```

Which is the right directory, but it doesn't seem to have worked

giorgiosironi:
```
elife@ci--search:/srv/search$ ls -dl var/
drwxrwsr-x 4 www-data www-data 4096 Dec 13 15:20 var/
elife@ci--search:/srv/search$ ls -dl var/cache/
drwxr-sr-x 51 www-data www-data 4096 Dec 13 15:20 var/cache/
elife@ci--search:/srv/search$ ls -dl var/cache/metadata/
drwxrwsr-x 2 www-data www-data 4096 Dec 13 15:20 var/cache/metadata/
```
The `rws` triplet is correct as the `setgid` (the `s`) will cause nested folders to inherit the group `www-data` and be writable by the `elife` user and not just by the `www-data` one. However, `var/cache/` seems to have been created without the writing permission, which points to a wrong umask?

giorgiosironi:
The `umask(002)` in the constructor of the class smells, I would move it to the entry points: `bin/console` and `web/bootstrap.php`. After that, delete `var/` and restart the build to see if it is created correctly.

stephenwf:
@giorgiosironi How do I ssh into a specific pull request using `bldr` to delete the var dir?

giorgiosironi:
All pull requests of this project are build on `search--ci`, so SSH in there and go to /srv/search.

giorgiosironi:
Having problems on `search--end2end` with `/srv/search/cache/` so I'll try to get this to work to solve it.

giorgiosironi:
We were missing https://github.com/elifesciences/search-formula/commit/1b138640c182f6efbdb037cb941092e12b759b65 afterward fixed with https://github.com/elifesciences/search-formula/commit/34d70f91e78d65edbec8922309de6b807034ec4d

PR #56, opened 2016-12-16
Skipping nulls representing invalid articles

giorgiosironi:
Temporary hack to allow imports without a fully valid article corpus


giorgiosironi:
This doesn't seem to break havoc, so I'll try it out on `end2end` and then `prod`.

PR #59, opened 2016-12-19
Logging Gearman and supporting CLI

giorgiosironi:


src/Search/Annotation/GearmanTaskDriver.php
```diff
@@ -58,16 +60,17 @@ public function registerWorkflow(Workflow $workflow)
...
+            $this->addTaskToWorker($task, $worker);
         }
     }
 
-    public function addTaskToWorker(GearmanTaskInstance $task, GearmanWorker $worker, LoggerInterface $logger)
+    public function addTaskToWorker(GearmanTaskInstance $task, GearmanWorker $worker)
     {
-        $worker->addFunction($task->name, Closure::bind(function (GearmanJob $job) use ($task, $logger) {
+        $worker->addFunction($task->name, Closure::bind(function (GearmanJob $job) use ($task) {
+            $this->logger->info('GearmanTaskDriver task started', ['task' => $task->name]);
```
stephenwf: Under normal circumstances this part of the code should not fail at run time (memory issues aside) the errors should be caught in the workflow methods that are called lines 82/84.

Theres no real way to get accurate id/type without knowing more about which task is running and the format of the data/response. I think it would be best to keep that logic contained within the workflows. If there is an exception thrown in this code we could catch the exception, deserialize (using the workflow) and perform a best guess at the data, or just dump the raw data. That might help identify

stephenwf: Also the first task for any importing is always validating the json so that we can be assured that it will continue through the tasks without failing at that step, narrowing down the potential causes.

giorgiosironi: I'll turn down the levels here to `debug`, helps identifying malfunctioning Gearman or that no jobs are being added

src/Search/Annotation/GearmanTaskDriver.php
```diff
@@ -58,16 +60,17 @@ public function registerWorkflow(Workflow $workflow)
...
+            $this->addTaskToWorker($task, $worker);
         }
     }
 
-    public function addTaskToWorker(GearmanTaskInstance $task, GearmanWorker $worker, LoggerInterface $logger)
+    public function addTaskToWorker(GearmanTaskInstance $task, GearmanWorker $worker)
     {
-        $worker->addFunction($task->name, Closure::bind(function (GearmanJob $job) use ($task, $logger) {
+        $worker->addFunction($task->name, Closure::bind(function (GearmanJob $job) use ($task) {
+            $this->logger->debug('GearmanTaskDriver task started', ['task' => $task->name]);
```
giorgiosironi: agreed, adding it

stephenwf:
Much cleaner than passing the logging and output instance around 👍

src/Search/Kernel.php
```diff
@@ -177,11 +178,13 @@ public function dependencies(Application $app)
             $logger = new Logger('search-api');
             if ($app['config']['file_log_path']) {
                 $stream = new StreamHandler($app['config']['file_log_path'], Logger::DEBUG);
+                $stream->pushProcessor(new ProcessIdProcessor());
                 $stream->setFormatter(new JsonFormatter());
                 $logger->pushHandler($stream);
             }
             if ($app['config']['file_error_log_path']) {
                 $stream = new StreamHandler($app['config']['file_error_log_path'], Logger::ERROR);
+                $stream->pushProcessor(new ProcessIdProcessor());
```
stephenwf: 👍  I do like these monolog extensions

PR #62, opened 2016-12-19
Error fix

stephenwf:
@giorgiosironi One more change!

### DIFF
--- a/src/Search/Api/Elasticsearch/Command/BuildIndexCommand.php
+++ b/src/Search/Api/Elasticsearch/Command/BuildIndexCommand.php
@@ -60,7 +60,7 @@ class BuildIndexCommand extends Command
         $create = null;
 
         // Try removing old one.
-        if ($toDelete) {
+        if ($toDelete && $this->client->indexExists()) {
             try {
                 $delete = $this->client->deleteIndex();
             } catch (Throwable $e) {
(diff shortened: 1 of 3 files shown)


src/Search/Api/Elasticsearch/Command/BuildIndexCommand.php
```diff
@@ -67,7 +67,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             try {
                 $delete = $this->client->deleteIndex();
             } catch (Throwable $e) {
-                $logger->debug($e->getMessage(), $e->getTrace());
```
giorgiosironi: > The issue I've seen is on a clean environment where there is no index in the first instance, and an error is displayed saying "no index to delete"

Agreed that this shouldn't cause an error, but other deletion errors are still interesting to see. Therefore some possible ways to make it work are:
- checking if an index esists first, and deleting it only if it exists
- or, always trying to delete it, but logging with `info` instead of `error` in case the exception is exactly what we expect to happen (checking its class or message); other exceptions should still be errors

stephenwf: I tried the first instance, and for some reason I was getting NULL back from elasticsearch, I can re-look into that, it would be the easiest. I also tried just doing an inspect/get on the index, but the size of the response was massive!

giorgiosironi: The guy debugging his failed PR build months from now will thank you :)

stephenwf:
Just rebased, only showing one commit for some reason!

stephenwf:
Just rebased.. seems to have misplaced the previous change! I.. don't even know how

src/Search/Api/Elasticsearch/Command/BuildIndexCommand.php
```diff
@@ -60,27 +60,32 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-        }
-        if (isset($create['error'])) {
-            $this->logger->error('Index '.$create['error']['reason'].' skipping creation.');
+
+        if (!$this->client->indexExists()) {
+            // Try adding new one!
+            try {
+                $create = $this->client->customIndex($config);
+            } catch (Throwable $e) {
+                $this->logger->error($e->getMessage(), $e->getTrace());
```
giorgiosironi: should probably re-throw after catching and logging?

giorgiosironi: othwerise further `if`s like the one on `$create` may fail

stephenwf:
@giorgiosironi Added the rethrow

giorgiosironi:
Looks like something is crashing and preventing the import from completing?

stephenwf:
Since the last successful build there was nothing changed in the import script. Maybe an external service not working?

giorgiosironi:
`ci` should not use external services, I will debug this after building the other PRs.

PR #63, opened 2016-12-19
Added enqueue cli

stephenwf:
@giorgiosironi

### DIFF
--- a/src/Search/Console.php
+++ b/src/Search/Console.php
@@ -8,6 +8,8 @@ use eLife\Search\Api\Elasticsearch\ElasticsearchClient;
 use eLife\Search\Api\Elasticsearch\Response\DocumentResponse;
 use eLife\Search\Api\Elasticsearch\Response\SuccessResponse;
 use eLife\Search\Api\Response\BlogArticleResponse;
+use eLife\Search\Queue\Mock\QueueItemMock;
+use eLife\Search\Queue\WatchableQueue;
 use Exception;
 use GuzzleHttp\Client;
 use LogicException;
@@ -19,6 +21,7 @@ use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
 use Symfony\Component\Console\Question\Question;
 use Symfony\Component\Process\Process;
 use ZipArchive;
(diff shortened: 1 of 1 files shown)


stephenwf:
screen shot 2016-12-19 at 12 26 14

giorgiosironi:
Right, so since this uses the queue it rightly skips the `/articles` call and the `queue:watch` will request `/articles/:id` and import it.

src/Search/Console.php
```diff
@@ -54,6 +58,28 @@
...
+            'research-article',
+            'blog-articles',
+            'collection',
+        ], 0);
+        $type = $helper->ask($input, $output, $choice);
+        // Ge the Id.
+        $choice = new Question('<question>Whats the ID of the item to import: </question>');
+        $id = $helper->ask($input, $output, $choice);
+        // Create queue item.
+        $item = new QueueItemMock($type, $id);
```
giorgiosironi: given we are now using this for real items, `Mock` can be misleading; the other implementation is `SqsMessage` so they could become `BusSqsMessage` and `InternalSqsMessage` respectively?

src/Search/Console.php
```diff
@@ -54,6 +58,28 @@
...
+        $helper = new QuestionHelper();
+        // Get the type.
+        $choice = new ChoiceQuestion('<question>Which type would you like to import</question>', [
+            'research-article',
+            'blog-articles',
+            'collection',
+        ], 0);
+        $type = $helper->ask($input, $output, $choice);
+        // Ge the Id.
+        $choice = new Question('<question>Whats the ID of the item to import: </question>');
```
giorgiosironi: being asked is useful because typically one has no idea about the possible options. It would be useful to also have a way of running this as a single command line, you can imagine having to run this many times and that will be faster to repeat a test

giorgiosironi:
Some comments, but should be dealt with with the next PR as this is needed on `prod` in the meantime.

giorgiosironi:
```
elife@prod--search:/srv/search$ bin/console --env=prod queue:push
Which type would you like to import
  [0] research-article
  [1] blog-articles
  [2] collection
 > 0
Whats the ID of the item to import: 10627
```
results in
```
[2016-12-19 13:47:20] search-api.ERROR: There was an unknown problem importing research-article (10627) {"exception":"[object] (LogicException(code: 0): ApiSDK does not exist for the type `research-article`. at /srv/search/src/Search/Queue/BasicTransformer.php:48)","item":"[object] (eLife\\Search\\Queue\\SqsMessage: {})"} 
```

stephenwf:
Just pushed a fix:
https://github.com/elifesciences/search/pull/64

PR #66, opened 2016-12-19
Added new stage field

stephenwf:

### DIFF
--- a/src/Search/Api/Response/ArticleResponse/Article.php
+++ b/src/Search/Api/Response/ArticleResponse/Article.php
@@ -38,6 +38,13 @@ trait Article
     public $version;
 
     /**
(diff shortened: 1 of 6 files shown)


giorgiosironi:
Going through https://github.com/elifesciences/api-raml/commits/develop/src/snippets/article.v1.yaml there is also a `versionDate` being added.

stephenwf:
I've added that field too, looks to be optional though. Also added in validation when we are running through our tests, this will catch these in future. I do remember it being in before, but it is now

giorgiosironi:
Ok, I guess then updating the api-raml and running a build will catch similar errors?
Bringing this to production.

stephenwf:
Yes thats right, it won't catch optional fields, but won't cause BC breaks

PR #67, opened 2016-12-20
BatchingHttpClient should group requests in batches of 10

giorgiosironi:


src/Search/Kernel.php
```diff
@@ -220,8 +221,11 @@ public function dependencies(Application $app)
 
         $app['api.sdk'] = function (Application $app) {
             return new ApiSdk(
-                new Guzzle6HttpClient(
-                    $app['guzzle']
+                new BatchingHttpClient(
+                    new Guzzle6HttpClient(
+                        $app['guzzle']
+                    ),
+                    10
```
thewilkybarkid: Can this be configurable? We might want more in production to speed it up.

thewilkybarkid:
👍

src/Search/Kernel.php
giorgiosironi: yes, I'll follow up by putting it in the `config/` files

PR #70, opened 2016-12-21
Memory limit for gearman:worker

giorgiosironi:
This is already in place for QueueCommand, but I suggest it to unify those to use some `*Limit` objects and making them configurable in a single place.


giorgiosironi:
I'll follow up on this by applying the same pattern to the `queue:watch` processes.

PR #71, opened 2017-01-03
Fixed sorting and added options

stephenwf:

I've re-added the sorting, I'm still waiting to confirm this is working, having some issues getting up and running locally, I think the "build index" command may be faulty

### DIFF
--- a/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
+++ b/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
@@ -43,7 +43,7 @@ final class ElasticQueryBuilder implements QueryBuilder
     private function sort($sort = [])
     {
         $this->query['sort'] = $this->query['sort'] ?? [];
-        $this->query['sort'] = array_merge($this->query['sort'], $sort);
+        $this->query['sort'][] = $sort;
     }
 
     private function getSort($reverse = false)
@@ -66,7 +66,7 @@ final class ElasticQueryBuilder implements QueryBuilder
         $this->query['body']['query']['bool']['must'][] = $query;
     }
 
-    public function searchFor(string $string) : QueryBuilder
+    public function searchFor(string $string): QueryBuilder
     {
         if ($string !== '') {
             $this->query('match', ['_all' => $string]);
@@ -75,14 +75,14 @@ final class ElasticQueryBuilder implements QueryBuilder
         return $this;
     }
 
-    public function order(string $direction = 'desc') : QueryBuilder
+    public function order(string $direction = 'desc'): QueryBuilder
     {
         $this->order = $direction === 'desc' ? 'desc' : 'asc';
 
         return $this;
     }
 
-    public function paginate(int $page = 1, int $perPage = 10) : QueryBuilder
+    public function paginate(int $page = 1, int $perPage = 10): QueryBuilder
     {
         $this->query['from'] = ($page - 1) * $perPage;
         $this->query['size'] = $perPage;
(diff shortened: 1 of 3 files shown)


src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
```diff
@@ -66,47 +66,45 @@ private function must($query)
...
+        $this->sort("_score:{$this->getSort($reverse)}");
 
         return $this;
     }
 
-    public function sortByDate($reverse = false) : QueryBuilder
+    public function sortByDate($reverse = false): QueryBuilder
     {
-        $this->sort(['published' => $this->getSort($reverse)]);
+        $this->sort("published:{$this->getSort($reverse)}");
```
thewilkybarkid: Is this going to sort on a property called 'published'? If so, it's a bit more complicated that that. Collections don't have a 'published' property, instead has 'updated'. Articles do have a 'published' property, but we actually want to sort on 'statusDate'.

I think it needs to index a property called 'sortDate' or something, and populate that for each of the content types in the appropriate manner.

stephenwf: I see, I will add a ticket into JIRA for that

stephenwf: For now I've added the sorts in the order:

- statusDate (ignored if not there)
- published
- updated (fallback if no published)

So this will hopefully resolve the issue in the short term until the ticket is completed.

thewilkybarkid: That compares them together, rather than sequentially?

stephenwf: I think as long as PHP keeps the order that is there it should do them sequentially

https://www.elastic.co/guide/en/elasticsearch/guide/current/_sorting.html#_multilevel_sorting

thewilkybarkid: Ok, that's not right (need to be compared together, otherwise articles will always appear first) but go with it for the moment (main list in the homepage will only be articles).

giorgiosironi:
The build is passing so this isn't breaking the indexing of api-dummy that we do in `ci`. However there is no test coverage for this feature, what would you like to add? Are PHPUnit tests on fake data all we need?

stephenwf:
I'm not entirely sure what to test, we are declaratively building a query and then passing it straight to ElasticSearch, the result sets themselves don't change behaviour.

If we do go down the route of testing the controller, we could provide some regression tests to say "given the conditions X expect the ES query to be Y". The value of such a test though is not really clear, since the domain itself is the interface. 

To test the implementation is correct against the interface we would need to have an ElasticSearch in the tests that we populate with controlled data and make sure our rules work. Unfortunately ES is un-deterministic for testing in some cases.

Alternatively we could set up something on the CI that will make assertions against the running site, in a more end-to-end fashion.

Any ideas or feedback would be great.

giorgiosironi:
Are we able to insert controlled data in ElasticSearch during the `ci` build (whatever is inside `project_tests.sh`)?

If so, what stop us from writing a test like this?
- (delete dirty data if necessary)
- insert document A with sorted field having value 1
- insert document B with sorted field having value 2
- query and checks B is the first returned result

stephenwf:
If the unit tests run after elasticsearch is bootstrapped and ready to handle results then we could contain it in a `WebTestCase` in the unit tests. This would require some tooling to allow inserting test information into the elasticsearch instance and deleting it.

stephenwf:
@giorgiosironi @thewilkybarkid I've added issue ELPP-1995 to cover the piece of work for adding the web test case. I will sit down with @elrossco22 once he's back and review other sections of the code we could cover with this approach and deliver as a single piece.

giorgiosironi:
Tests run after the formula, so they should find an ElasticSearch instance running there (pending the fact that it doesn't take 1 minute to start it, but if it's slow we should wait in the formula before declaring it complete.)

As for deleting data, can we use a console command that drops/recreates the index? We should be robust to tests leaving leftovers around.

PR #74, opened 2017-01-05
Added dataSet and funding to serialized set

stephenwf:
After this I can import ~3000 articles, with 200 missing from pages and a few other errors in the gist attached to the other ticket. This should result in a complete data set once those few date times are fixed on the articles endpoint.


giorgiosironi:
I suppose these are relatively new fields, :+1:

PR #75, opened 2017-01-05
Web test [wip]

stephenwf:
@giorgiosironi 

I will rework this as an base class for actual tests, but this is the basics of it working with console and app.

Next steps will be to set up an elasticsearch example. I'm thinking of changing the configuration (elasticsearch database name) and then just deleting the index afterwards.

### DIFF
--- a/tests/src/Search/ExampleWebTest.php
+++ b/tests/src/Search/ExampleWebTest.php
@@ -21,6 +21,8 @@ class ExampleWebTest extends WebTestCase
     protected $console;
     /** @var Kernel */
     protected $kernel;
+    /** @var ElasticsearchClient */
+    private $client;
 
     /**
      * Creates the application.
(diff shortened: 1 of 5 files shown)


giorgiosironi:
Nice to be able to perform fake HTTP requests and to execute commands (at least the ones not depending on SQS). A separate database makes sense not to pollute the main one used by the import, for example, but in general tests and import won't be run at the same time so if each cleans ElasticSearch before starting there would be no interaction problem.

tests/src/Search/ExampleWebTest.php
```diff
@@ -0,0 +1,153 @@
...
+        $api = $this->createClient();
+        $api->request('GET', '/search');
+        $response = $api->getResponse();
+        $this->assertTrue($response->isOk());
+        $json = json_decode($response->getContent());
+
+        $this->assertEquals($json->total, 1);
+        $this->assertEquals($json->items[0]->status, 'vor');
+        // ...
+        $client->deleteIndex();
```
giorgiosironi: this should go into a tearDown() method as it should be executed regardless of previous failures

giorgiosironi:
:+1:

PR #73, opened 2017-01-04
Instrumenting ci-import to improve logging of failures

giorgiosironi:
I have seen the import failing a few times in pull requests, but the only thing that we check is that the number of Gearman jobs goes to zero (no matter if they are discarded as failed). The goal is to make the build read if at least a Gearman job fails.

PR #76, opened 2017-01-06
New Relic for long-running processes

giorgiosironi:
Pull request that depends on https://github.com/elifesciences/search/pull/73
Adds monitoring of every single Gearman task being executed, or element being accepted from SQS.
Renames the import command to `queue:import` as it does not have anything to do with Gearman anymore.

### DIFF
--- a/bin/ci-import
+++ b/bin/ci-import
@@ -20,7 +20,7 @@ echo "Starting a queue:watch"
 ./bin/console queue:watch --env="$environment" 2>&1 | tee /tmp/queue-watch.log &
 queueWatchPid=$!
 echo "Starting import"
-./bin/console gearman:import all --env="$environment"
+./bin/console queue:import all --env="$environment"
 
 # will return also on failing conditions such as dead processes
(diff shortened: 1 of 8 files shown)


stephenwf:
Had a look through, looks good. We have a centralised task runner, do you think it would be useful to have a custom metric set up in the "GearmanTaskDriver" for duration of running each task. Would give us some nice graphs on NR side

stephenwf:
Possibly even hooking it up to Guzzle and having metrics for our external API calls for the long running processes too might be useful for debugging.

giorgiosironi:
As far as I know the New Relic PHP extension measures external HTTP calls automatically (hooking in curl_*) and also measure time for single transactions (that's why we mark them with `newrelic_start_transaction()` and `newrelic_end_transaction()`)

stephenwf:
Ah good to know, I didn't realise that sort of thing was just bundled in

giorgiosironi:
I'll merge but after #73, because it depends on it.

PR #77, opened 2017-01-06
Sorting fixes + webtest

stephenwf:


src/Search/Console.php
```diff
@@ -120,6 +120,10 @@ public function __construct(Application $console, Kernel $app)
         $this->app = $app;
         $this->root = __DIR__.'/../..';
 
+        if (!defined('GEARMAN_INSTALLED')) {
```
giorgiosironi: this should never be false. Where it is required? Should be made explicit in the composer.json as `ext-gearman`

tests/src/Search/ElasticTestCase.php
```diff
@@ -0,0 +1,158 @@
...
+        return $fn ? $fn($this->kernel->get('elastic.client')) : $this->kernel->get('elastic.client');
+    }
+
+    public function setUp()
+    {
+        parent::setUp();
+        $this->client = $this->getElasticSearchClient();
+        $lines = $this->runCommand('search:setup');
+        if ($lines[0] === 'No alive nodes found in your cluster') {
+            $this->markTestSkipped('Elasticsearch may not be installed, skipping');
```
giorgiosironi: I see this as a potential problem being ignored, given we will run this with a separate command, can we go for `$this->fail()` instead?

tests/src/Search/ElasticTestCase.php
```diff
@@ -0,0 +1,158 @@
...
+
+    public function runCommand(string $command)
+    {
+        $log = $this->returnCallback(function ($message) use (&$logs) {
+            $logs[] = $message;
+        });
+        $logs = [];
+        $logger = $this->createMock(NullLogger::class);
+
+        foreach (['debug', 'info', 'alert', 'notice', 'error'] as $level) {
```
giorgiosironi: missing `warning` and `critical` that we are using, plus `emergency` and similar, go for the full list

tests/src/Search/ExampleWebTest.php
```diff
@@ -0,0 +1,73 @@
...
+        $this->assertEquals('Hello from the outside (of the global scope)', $lines[0]);
+        $this->assertEquals('This is working', $lines[1]);
+    }
+
+    /**
+     * @test
+     */
+    public function testElasticSearchIndex()
+    {
+        $this->addDocumentToElasticSearch('
```
giorgiosironi: would produce better error messages to perform `json_encode()` over a PHP array, because a syntax error would be difficult to catch in the current state

src/Search/Console.php
stephenwf: Ah this is to aid local development. For example I have 2 PHP instances, a clean one without XDebug and Gearman for running through the tests in > 2s and another with XDebug and Gearman for other tests.

Its found in the test bootstrap.php but does not get defined when testing console commands, it must run it in isolation.

tests/src/Search/ElasticTestCase.php
stephenwf: Was intended to aid development, we should have caught any errors during the salt provisioning, I could change it to risky? Would that fail the build?

src/Search/Console.php
giorgiosironi: I agree with keeping xdebug out, and also the Gearman daemon, but the extension is a set of classes, it shouldn't create a performance problem?

stephenwf: We could change how it works, but currently when the Console class is constructed it creates all of the other commands, which in turn creates a Gearman client and looks for a Gearman daemon, the only drawback is a PHP notice saying "no gearman instances found" but I thought I should detect its existence and not construct gearman if its not installed. Maybe it would be better to have gearman be "lazy"

tests/src/Search/ElasticTestCase.php
giorgiosironi: I didn't find a way to mark a test as risky explicitly, let's keep it and I will specify later in `phpunit.xml` to exclude this groups of tests by default (but to always run them in CI)

src/Search/Console.php
giorgiosironi: There is a similar problem with SQS, on one hand it's useful to have things running with the least amount of dependencies and on the other hand it's possible to mask errors instead of discovering them sooner in `ci`, `end2end` and `prod`. Let's keep this for now.

tests/src/Search/ElasticTestCase.php
stephenwf: I've changed it to a fail for now, if thats also works? I've got these tests in a "web" group so locally we can do `phpunit --exclude-group=web`

giorgiosironi:
Yes, `phpunit --exclude-group=web` also works, what I want to do later is just to have the default `vendor/bin/phpunit` command exclude that group by tweaking `phpunit.xml` for convenience of typing a shorter command.

giorgiosironi:
Re-running the build as it got run while I was tweaking Gearman+Postgresql.

giorgiosironi:
Smoke test failure, because the index is not present. Maybe functional tests should in general delete the index, but recreate it immediately? It's like `DELETE FROM` instead of `DROP TABLE` for database fixtures.

stephenwf:
Do the smoke tests still run after the unit tests? I think this will cause it to break

giorgiosironi:
(yes, they do, and they break)

stephenwf:
Would we need to re-import content after running the tests too for the smoke test?

giorgiosironi:
No, they are explicitly built to run on all environments so they make no assumptions on the content: just that the status code is not 500.

stephenwf:
I've changed the index for when we run the tests, this should ensure the tests are isolated, worked fine locally

giorgiosironi:
Non deterministic assertion failure... I think you need `assertEquals` instead of `assertSame` as JSON does not preserve the order of fields?

PR #78, opened 2017-01-09
Performing ci-import explicitly in CI build

giorgiosironi:
Will avoid doing this in the formula, so that we can customize the order and perform deploy+tests+import in this order, not (deploy and import)+tests


giorgiosironi:
Just waiting for the previous merged PR to reach production, then I'll merge.

PR #79, opened 2017-01-09
Sigterm

stephenwf:
@giorgiosironi


src/Search/Annotation/GearmanTaskDriver.php
```diff
@@ -133,7 +137,7 @@ public function work()
         $this->logger->info('Worker started.');
         $this->addTasksToWorker($this->worker);
         $limit = $this->limit;
-        while (!$limit()) {
+        while (!$limit() && Signals::isValid()) {
```
giorgiosironi: can you unify `Signals` with `MemoryLimit` in a Composite pattern so that here we can just do:
```
while ($limit()) {
    ...
}
```
where that object being injected and created in Kernel is:
```
new CompositeLimit([
    MemoryLimit::mb(...),
    SignalsLimit::sigterm(),
    ... // any other reason for stopping we add later
]);
```

src/Search/Gearman/Command/QueueCommand.php
```diff
@@ -47,6 +48,9 @@ public function __construct(
         $this->logger = $logger;
         $this->monitoring = $monitoring;
         $this->limit = $limit;
+        // Signals.
+        Signals::register();
```
giorgiosironi: if `Signals` is an object we can put the `->register()` call in the `bin/console` file like for https://github.com/elifesciences/search/blob/develop/bin/console#L30
which would make this class less dependent on it, and more testable as the constructor does not perform any special work when used inside a test case

src/Search/Annotation/GearmanTaskDriver.php
giorgiosironi: can you unify `Signals` with `MemoryLimit` in a Composite pattern so that here we can just do:
```
while ($limit()) {
    ...
}
```
where that object being injected and created in Kernel is:
```
new CompositeLimit([
    MemoryLimit::mb(...),
    SignalsLimit::sigterm(),
    ... // any other reason for stopping we add later
]);
```

src/Search/Annotation/GearmanTaskDriver.php
```diff
@@ -150,6 +154,8 @@ public function work()
 
                 return;
             }
+            // Dispatch any outstanding signals.
+            Signals::tick();
```
giorgiosironi: To keep a small interface, this work could be done in `Signals::isValid()` (no matter how it is renamed to uniform interfaces)

src/Search/Gearman/Command/ApiSdkCommand.php
```diff
@@ -24,11 +26,11 @@
     private $output;
     private $logger;
     private $monitoring;
+    private $queue;
 
     public function __construct(
         ApiSdk $sdk,
-        // TODO: type hint
-        $queue,
+        WatchableQueue $queue,
```
giorgiosironi: :+1:

src/Search/Gearman/Command/QueueCommand.php
```diff
@@ -91,9 +95,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
         // Loop.
         $limit = $this->limit;
         while (!$limit()) {
```
giorgiosironi: I think the conditions in commands can be:
```
while (!($reason = $limit())) {
    ...
}
```
where `$reason` is a string description of the reason for stopping. Then we can use it in the `$this->logger->info()` at the end of the method to say `Stopped because ...`

src/Search/Signals.php
```diff
@@ -0,0 +1,34 @@
...
+class Signals
+{
+    private static $isValid = true;
+
+    public static function register()
+    {
+        if (function_exists('pcntl_signal')) {
+            // Signals
+            pcntl_signal(SIGTERM, [self::class, 'onTermination']);
+            pcntl_signal(SIGHUP, [self::class, 'onTermination']);
```
giorgiosironi: we may want to add `SIGINT` to deal with `Ctrl+C`

src/Search/Signals.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace eLife\Search;
+
+class Signals
+{
+    private static $isValid = true;
+
+    public static function register()
+    {
+        if (function_exists('pcntl_signal')) {
```
giorgiosironi: when is this not present? I can think of Windows for example, but I'd like a clear error message rather than a no-op as it's important functionality

src/Search/Signals.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace eLife\Search;
+
+class Signals
+{
+    private static $isValid = true;
```
giorgiosironi: As for previous comments, an object with a Factory Method in Kernel and injected in the right places would be more amenable to polymorphism

stephenwf:
@giorgiosironi I've added a SignalLimit and CompositeLimit, I can remove the Signal class if you're happy with the implementation

The PCTL extension is again missing locally, I can remove the guards, use the logger, or throw some error though if thats preferable.

giorgiosironi:
In the build: https://ci--alfred.elifesciences.org/job/pull-requests-projects/job/search/job/PR-79/9/console
it appears that the check failing is on the return code of (annotated):
```
# (launches the "waiting" process, but in background)
timeout 60 ./bin/wait-for-empty-gearman-queue "$environment" & 
# captures the pid of "waiting"
waitingPid=$!
# waits for it to finish, which it did in 10 seconds
wait "$waitingPid"
# checks the return code of `wait`, which should be the one of the command originally being run, wait-for-empty-gearman-queue
waitingReturnCode=$?
```
So the script that was waiting for the queues has found an error.
This:
```
12:10:38 [52.87.244.171] out: bin/ci-import: line 31: kill: (12192) - No such process
12:10:38 [52.87.244.171] out: bin/ci-import: line 31: kill: (12194) - No such process
```
points to the processes having crashed, which makes the waiting stop because with crashed processes you cannot empty the queue. So I would try to execute manually the bin/console processes to check what happens to them.

bin/ci-import
```diff
@@ -28,7 +28,7 @@ waitingPid=$!
 wait "$waitingPid"
 waitingReturnCode=$?
 # clean up PHP processes
-kill "$workerPid" "$queueWatchPid";
+kill "$workerPid" "$queueWatchPid" || true;
 if [ ! "$waitingReturnCode" -eq "0" ]; then
     echo "Waiting for queues to be empty timed out"
```
giorgiosironi: we should probably print `$waitingReturnCode` here

giorgiosironi: and also substitute the echo with `echo "./wait-for-empty-gearman-queue timed out" to be more clear`

stephenwf:
Here is what I got from the logs on the watch command
```
{"message":"queue:watch: Started listening.","context":[],"level":200,"level_name":"INFO","channel":"search-api","datetime":{"date":"2017-01-09 15:50:12.570017","timezone_type":3,"timezone":"UTC"},"extra":{"process_id":4713}}
{"message":"queue:watch: Stopped because of limits reached.","context":[],"level":200,"level_name":"INFO","channel":"search-api","datetime":{"date":"2017-01-09 15:50:12.570186","timezone_type":3,"timezone":"UTC"},"extra":{"process_id":4713}}
```

Ill take a look at my logic for the limits again

bin/ci-import
```diff
@@ -28,9 +28,9 @@ waitingPid=$!
 wait "$waitingPid"
 waitingReturnCode=$?
 # clean up PHP processes
-kill "$workerPid" "$queueWatchPid";
+kill "$workerPid" "$queueWatchPid" || true;
 if [ ! "$waitingReturnCode" -eq "0" ]; then
-    echo "Waiting for queues to be empty timed out"
+    echo -e "\e[31mThe script './wait-for-empty-gearman-queue' timed out\e[39m"
```
giorgiosironi: ok, ANSI colors are supported also be Jenkins

bin/ci-import
```diff
@@ -28,9 +28,9 @@ waitingPid=$!
 wait "$waitingPid"
 waitingReturnCode=$?
 # clean up PHP processes
-kill "$workerPid" "$queueWatchPid";
+kill "$workerPid" "$queueWatchPid" || true;
```
giorgiosironi: this is a smell, they shouldn't terminate by themselves as they are long-running processes

src/Search/Kernel.php
```diff
@@ -202,7 +205,11 @@ public function dependencies(Application $app)
         };
 
         $app['limit'] = function (Application $app) {
-            return MemoryLimit::mb($app['config']['process_memory_limit']);
+            return new LoggingMiddleware(
+                new CompositeLimit(
+                    MemoryLimit::mb($app['config']['process_memory_limit']),
+                    SignalsLimit::sigterm(['SIGINT', 'SIGTERM', 'SIGHUP'])
```
giorgiosironi: method name and parameters are in disagreement about which signals are captured?

src/Search/Limit/CompositeLimit.php
```diff
@@ -0,0 +1,33 @@
...
+    public $functions = [];
+
+    public function __construct(Limit ...$args)
+    {
+        $this->functions = $args;
+    }
+
+    public function __invoke(): bool
+    {
+        $failures = array_filter($this->functions, function (Limit $fn) {
```
giorgiosironi: wasn't a foreach() easier than a side-effect inside filtering? :thinking:

src/Search/Limit/MemoryLimit.php
```diff
@@ -24,4 +24,9 @@ public function __invoke() : bool
 
         return false;
     }
+
+    public function getReasons() : array
+    {
```
giorgiosironi: Thinking about it, the original could be improved if it saved in $this->actualBytes the result of `memory_get_usage(true)`. That way we would be able to see if there are big jumps of memory like `Memory limit exceeded: 200000000 exceedes 100000000`

src/Search/Limit/SignalsLimit.php
```diff
@@ -0,0 +1,43 @@
...
+    public function __invoke() : bool
+    {
+        pcntl_signal_dispatch();
+
+        return $this->valid === false;
+    }
+
+    public function getReasons(): array
+    {
+        return ['Signals: Stopped by user'];
```
giorgiosironi: also here it can be helpful for debugging to trace which was the signal

giorgiosironi: `pcntl_signal(self::$validSignals[$signal], function() use ($signal) { $this->onTermination[$signal]; });`

tests/src/Search/LimitTest.php
```diff
@@ -0,0 +1,74 @@
...
+        $fail = new BasicLimitMock(true);
+        $fail2 = new BasicLimitMock(true);
+        $fail3 = new BasicLimitMock(true);
+        $pass = new BasicLimitMock();
+        $limit = new CompositeLimit($fail, $fail2, $fail3, $pass);
+
+        $this->assertTrue($limit());
+
+        $this->assertEquals([
+            'This is the reason it failed',
```
giorgiosironi: would be a more precise test to configure different messages

PR #81, opened 2017-01-10
Updating api-sdk to newest version, without caching

giorgiosironi:


stephenwf:
Looks good, do we have any metrics to compare against to show how this will affect memory during importing or queue watching?

giorgiosironi:
Memory usage: https://rpm.newrelic.com/accounts/1451451/servers/26879341
Speed: https://ci--alfred.elifesciences.org/job/test-search/ (building times in ci/end2end, small dataset)

stephenwf:
End to end times have jumped significantly!

giorgiosironi:
Eheh no that's https://github.com/elifesciences/search/commit/b86e0555823286a95b948272fde306c31f424360
Previously search was running just a couple of end2end tests, but now it's been uniformed to the other projects in running all of them.

PR #83, opened 2017-01-10
gearman:worker should not hang on work()

giorgiosironi:
Like for queue:watch, it should wait for a finite amount and then loop to check for signals or other limits. The default behavior seem to be an infinite timeout.

### DIFF
--- a/bin/ci-import
+++ b/bin/ci-import
@@ -14,10 +14,10 @@ trap 'kill -INT -$waitingPid' INT
 sleep 5
 
 echo "Starting a gearman:worker"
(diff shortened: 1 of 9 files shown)


giorgiosironi:
Going with this otherwise all builds fail, but feel free to add comments.

stephenwf:
The limits are not errors, but they are the cause of stopping the execution

giorgiosironi:
Point was: when you log in into a server and check what's wrong in the `error.log`, you want to see things that have gone wrong, but a deployment would put the restart of the processes in there even if it's expected

stephenwf:
That makes sense

PR #84, opened 2017-01-10
Dead code: release() method

giorgiosironi:
This is not used at the moment, and in general we have started to log errors and remove the message from the queue rather than letting it there for a possibly infinite number of retries

### DIFF
--- a/tests/src/Search/ExampleWebTest.php
+++ b/tests/src/Search/ExampleWebTest.php
@@ -12,11 +12,8 @@ class ExampleWebTest extends ElasticTestCase
      */
     public function testCanRunCommand()
     {
-        // Run command
-        $lines = $this->runCommand('hello');
-        // And test.
-        $this->assertEquals('Hello from the outside (of the global scope)', $lines[0]);
-        $this->assertEquals('This is working', $lines[1]);
+        $logs = $this->runCommand('help');
+        $this->assertEquals(0, count($logs));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


stephenwf:
Yeh that makes more sense, is there a case where we need to release the items again? I suppose since we have persistence for Gearman it shouldn't matter

giorgiosironi:
We haven't encountered temporary errors yet, but if it happens we can avoid removing the element from the queue and it will be added back automatically at the end of the visibility timeout.

giorgiosironi:
There is also `Console::ramlCommand()`, given it is managed with Composer now should we remove that too?

stephenwf:
I think theres a few things in console that were there to speed up further development of commands from the eLife Silex Starter, such as `hello` `echo` `spawn` `debug:search:random` and `debug:search`

giorgiosironi:
Are they necessary now?

stephenwf:
No I don't think so

tests/src/Search/ExampleWebTest.php
```diff
@@ -12,11 +12,8 @@ class ExampleWebTest extends ElasticTestCase
      */
     public function testCanRunCommand()
     {
-        // Run command
-        $lines = $this->runCommand('hello');
-        // And test.
-        $this->assertEquals('Hello from the outside (of the global scope)', $lines[0]);
-        $this->assertEquals('This is working', $lines[1]);
+        $lines = $this->runCommand('help');
+        $this->assertGreaterThan(1, count($lines));
```
stephenwf: 👍  Forgot about that test that uses it, good catch

PR #85, opened 2017-01-10
User-Agent configuration

giorgiosironi:
It is starting to become useful to understand where a request comes from when debugging. All requests are usually mediated through the api-gateway so it is not immediate from ip addresses (that vary anyway very frequently).

### DIFF
--- a/src/Search/Kernel.php
+++ b/src/Search/Kernel.php
@@ -234,6 +234,9 @@ final class Kernel implements MinimalKernel
             return new Client([
                 'base_uri' => $app['config']['api_url'],
                 'handler' => $stack,
+                'headers' => [
+                    'User-Agent' => 'elifesciences/search',
+                ],
             ]);
         };
 


giorgiosironi:
@thewilkybarkid could be useful across the board

thewilkybarkid:
Yeah, been meaning to sort it. I'd like the API Client and API SDK to both add details, then the end users should do so too.

giorgiosironi:
Separate headers to avoid clashes?

thewilkybarkid:
`User-Agent` takes multiple details. Already it's something like `GuzzleHttp/6.2.1 curl/7.51.0 PHP/7.1.0` by default, so I'd expect it to be along the lines of `eLifeSearch/e437748e478bf43936b1d00c324f7c24a44b24e1 eLifeApiSdk/1.0.1 eLifeApiClient/1.0.0 GuzzleHttp/6.2.1 curl/7.51.0 PHP/7.1.0`.

giorgiosironi:
How should a project find out the User-Agent configured by the library it includes?

stephenwf:
We can put this is some helper method to "Add a user agent"

```php
$headers = $client->getDefaultOption('headers');
$headers['User-Agent'] .= ' eLifeSearch/{something}';
$client-> setDefaultOption('headers', $headers);
```

thewilkybarkid:
API-Client should do it in `Guzzle6HttpClient` (or any other implementation that we do), prepending to the default. API-SDK should do something in `ApiSdk` to cause its value to be prepended, along with an optional user value.

If someone sets the value in the Guzzle client directly, that would just override it.

PR #86, opened 2017-01-10
Adding a boatload of logs to find out the race condition in queue:watch

giorgiosironi:
The test of indexing a new article is randomly failing, so these will help understand whether any 404 is reused with respect to making a new request to /articles/:id.


stephenwf:
Maybe we can keep these behind the debug flag once we know whats wrong?

giorgiosironi:
We would probably remove most of them once fixed.

giorgiosironi:
(it has to run in `end2end` now)

PR #89, opened 2017-01-13
Differentiating limits into long_running and interactive

giorgiosironi:
Long-running processes are limited both on signals (deployments) and memory (they can be restarted by Upstart.)

Interactive processes instead are limited on signals (for interrupting them gracefully) but not on memory as they won't be restarted automatically, and have to go back to the start if they are.

### DIFF
--- a/src/Search/Kernel.php
+++ b/src/Search/Kernel.php
@@ -207,16 +207,16 @@ final class Kernel implements MinimalKernel
             return new Monitoring();
         };
 
-        /**
+        /*
          * @internal
          */
-        $app['limit._memory'] = function(Application $app) {
+        $app['limit._memory'] = function (Application $app) {
             return MemoryLimit::mb($app['config']['process_memory_limit']);
-        });
-        /**
+        };
+        /*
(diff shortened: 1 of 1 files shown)


src/Search/Kernel.php
```diff
@@ -207,12 +207,36 @@ public function dependencies(Application $app)
...
+                ),
+                $app['logger']
+            );
+        };
+
+        $app['limit.interactive'] = function (Application $app) {
+            return new LoggingMiddleware(
+                new CompositeLimit(
+                    $app['limit._signals']
+                ),
```
stephenwf: Probably don't need the composite limit here, just the `$app['limit._signals']`

giorgiosironi: True, I wanted the `LoggingMiddleware`, should I keep it?

stephenwf: Logging Middleware will work with a single limit, but if we add more limits it might be useful to keep this composite for clarity, they only get called on cli commands so I doubt the overhead is a concern

PR #92, opened 2017-01-17
ELPP-1929 Increased count accuracy

stephenwf:
@giorgiosironi @elrossco22 @thewilkybarkid


stephenwf:
We can possibly have a persisted value (apc maybe?) that counts with accuracy. We could also optimise empty result sets to not have aggregations and instead cache the aggregation counts.

As for the smoke tests, since this is still nondeterministic and could pass smoke tests while being broken, I think this would be best suited to manual testing. Its unlikely that these will break without a change in ES version or drastic change to queries. I can comment something alone these lines in the QueryBuilder class itself

PR #96, opened 2017-01-23
ELPP-1890 DateTime -> DateTimeImmutable

stephenwf:
@giorgiosironi @elrossco22 

Will be applying this to medium and recommendations shortly (needed to update dependencies here anyway)

### DIFF
--- a/tests/src/Search/Api/Response/VorArticleResponseTest.php
+++ b/tests/src/Search/Api/Response/VorArticleResponseTest.php
@@ -28,7 +28,6 @@ class VorArticleResponseTest extends SerializerTest
                     "version": 1,
                     "elocationId": "e09560",
                     "doi": "10.7554\/eLife.09560",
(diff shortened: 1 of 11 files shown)


tests/src/Search/Api/Response/VorArticleResponseTest.php
```diff
@@ -28,7 +28,6 @@ public function jsonProvider() : array
                     "version": 1,
                     "elocationId": "e09560",
                     "doi": "10.7554\/eLife.09560",
-                    "authorLine": "Lee R Berger et al",
```
giorgiosironi: not sure how this is related, but if it makes the test pass ok

stephenwf: I believe the authorLine field is now optional, so it doesn't appear in the minimal JSON (an update to the RAML changed the fixtures we test against)

PR #97, opened 2017-01-25
Fixed subjects being loaded at runtime

stephenwf:
@giorgiosironi @thewilkybarkid This should fix the issue during load tests with the external calls to CMS


thewilkybarkid:
Looks like the `SubjectStore` is used any more by the controller.

stephenwf:
@thewilkybarkid Removed that now, good to merge?

PR #98, opened 2017-01-25
ELPP-2160 Order of content on homepage

stephenwf:
@giorgiosironi @thewilkybarkid @elrossco22 

I've changed the way we are sorting to instead be a "virtual" property that we can choose at insert time what it is. This will increase performance at the cost of having to reindex if we decide to change which fields we use. There are 2 solutions I can see for solving this issue in the future:

- Robust solution for zero downtime migrations (reindex content into new index, switcheroo)
- Process for creating and applying migrations for changes to indexes

This does however solve the immediate issue.

### DIFF
--- a/src/Search/Api/Elasticsearch/Command/resources/labs-experiment.yaml
+++ b/src/Search/Api/Elasticsearch/Command/resources/labs-experiment.yaml
@@ -66,6 +66,3 @@ labs-experiment:
       type: string
     number:
       type: long
-    type:
-      type: string
-      index: not_analyzed
--- a/src/Search/Workflow/BlogArticleWorkflow.php
+++ b/src/Search/Workflow/BlogArticleWorkflow.php
@@ -20,6 +20,7 @@ final class BlogArticleWorkflow implements Workflow
     const WORKFLOW_FAILURE = -1;
 
     use JsonSerializeTransport;
+    use SortDate;
 
     /**
      * @var Serializer
@@ -84,8 +85,7 @@ final class BlogArticleWorkflow implements Workflow
         $this->logger->debug('BlogArticle<'.$blogArticle->getId().'> Indexing '.$blogArticle->getTitle());
         // Normalized fields.
         $blogArticleObject = json_decode($this->serialize($blogArticle));
-        $sortDate = $blogArticle->getPublishedDate();
-        $blogArticleObject->sortDate = $sortDate->format('Y-m-d\TH:i:s\Z');
+        $this->addSortDate($blogArticleObject, $blogArticle->getPublishedDate());
 
         // Return.
         return [
--- a/src/Search/Workflow/BlogArticleWorkflow.php
+++ b/src/Search/Workflow/BlogArticleWorkflow.php
@@ -81,11 +81,15 @@ final class BlogArticleWorkflow implements Workflow
      */
     public function index(BlogArticle $blogArticle) : array
     {
-        // This step is still not used very much.
         $this->logger->debug('BlogArticle<'.$blogArticle->getId().'> Indexing '.$blogArticle->getTitle());
+        // Normalized fields.
+        $blogArticleObject = json_decode($this->serialize($blogArticle));
+        $sortDate = $blogArticle->getPublishedDate();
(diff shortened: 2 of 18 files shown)


stephenwf:
@thewilkybarkid  I'm going to expand the ordering of content test to include more examples and varied content. Can you remind me when the `updated` field becomes the field to sort on for date

thewilkybarkid:
For articles (as in research articles, research advances etc) it's the `statusDate`. For collections it's the `updated` date. All other content types have `published` dates. (I think!)

stephenwf:
What about for events? They don't appear to have a published date, only a start and end date (which in the dummy-api is pushing all future events to the top!)

thewilkybarkid:
Ah yeah, would have to be `starts`.

I wonder if we should be indexing events at all... (and if so, if they should disappear once they've completed).

stephenwf:
Okay so thats passing the original test case now, although I cannot find the `updated` field on collections
The past events dropping off is probably out of scope for this particular issue.

src/Search/Api/Elasticsearch/Command/resources/labs-experiment.yaml
```diff
@@ -66,6 +66,3 @@ labs-experiment:
       type: string
     number:
       type: long
-    type:
```
giorgiosironi: removing the field because it's not used anymore?

src/Search/Workflow/BlogArticleWorkflow.php
```diff
@@ -81,11 +81,15 @@ public function validate(BlogArticle $blogArticle) : BlogArticle
      */
     public function index(BlogArticle $blogArticle) : array
     {
-        // This step is still not used very much.
         $this->logger->debug('BlogArticle<'.$blogArticle->getId().'> Indexing '.$blogArticle->getTitle());
+        // Normalized fields.
+        $blogArticleObject = json_decode($this->serialize($blogArticle));
+        $sortDate = $blogArticle->getPublishedDate();
+        $blogArticleObject->sortDate = $sortDate->format('Y-m-d\TH:i:s\Z');
```
giorgiosironi: this assignment to sortData should be extracted, the format for example is a detail that should be kept in one place. Method in a trait or abstract class?

giorgiosironi: (of course the `getPublishedDate()` vs other options changes in each content type, so it should not be extracted)

thewilkybarkid:
https://github.com/elifesciences/api-raml/blob/develop/src/snippets/collection.v1.yaml#L13-L15 is the Collection date.

Agree re events, keep them as is. We'll need to talk to our Product team to see if they want to change anything. Easier once we have a working set of data too.

src/Search/Workflow/BlogArticleWorkflow.php
thewilkybarkid: I wonder if that could go in the SDK.

thewilkybarkid: A method on `Model`. Or even a `DateSortedModel` or something.

stephenwf: I've split it out into a trait for now

stephenwf:
Hmm I can't see the pair of that field in the SDK:
https://github.com/elifesciences/api-sdk-php/blob/master/src/Model/Collection.php

In the SDK is it the published field that is the same as the updated?

src/Search/Api/Elasticsearch/Command/resources/labs-experiment.yaml
stephenwf: Yes its in the `_default_.yml` file which is shared across all the indexes, just removing to avoid it causing issues

thewilkybarkid:
Huh, the SDK calling JSON `updated` date `published` (https://github.com/elifesciences/api-sdk-php/blob/master/src/Serializer/CollectionNormalizer.php#L96).

stephenwf:
Good to go @thewilkybarkid @giorgiosironi ?

PR #100, opened 2017-01-30
ELPP-2194 Subject and content type numbers dont change

stephenwf:
@giorgiosironi This also contains some change about CI logging, unsure why that was not pushed.

This does work, although the performance needs to be taken into consideration.


giorgiosironi:
We'll load test again from the user's perspective (journal pages).

PR #101, opened 2017-01-30
Added test covering information from ELPP-1929

stephenwf:
@giorgiosironi Covers the case you mentioned in 1929 (I cannot replicate)

### DIFF
--- a/tests/src/Search/EventDateTest.php
+++ b/tests/src/Search/EventDateTest.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace tests\eLife\Search;
+
+/**
+ * @group web2
+ */
+class EventDateTest extends ElasticTestCase
+{
+    public function test_event_end_date_is_correct()
+    {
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Maybe there is a problem in the SDK?

giorgiosironi:
Yes, found it.

giorgiosironi:
https://github.com/elifesciences/api-sdk-php/pull/99

giorgiosironi:
Would close this as the problem isn't here?

giorgiosironi:
Or maybe substitute with upgrade of api-sdk when fixed.

stephenwf:
I will wait until there is an update to `api-sdk`, this was meant to replicate the issue to try and track it down.

giorgiosironi:
Ok, but this test does not replicate it (probably it skips that part of the sdk in its data import phase) so a different test would be needed.

giorgiosironi:
Closing in favor of #102

PR #103, opened 2017-02-02
ELPP-2093 Added datetime filter to search api

stephenwf:
I've added in the date range filter now conforming to the RAML.

This is the index-less and unoptimised version, but it is working. We can see if it causes any slowdown on the archive pages and review adding month/year fields to each doc and doing a simple lookup

### DIFF
--- a/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
+++ b/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
@@ -155,7 +155,7 @@ final class ElasticQueryBuilder implements QueryBuilder
             'sortDate' => [
                 'gte' => $fromDate->format('Y/m/d'),
                 'lt' => $toDate->format('Y/m/d'),
-                'format' => 'yyyy/MM/dd||yyyy/MM/dd',
+                'format' => 'yyyy/MM/dd',
             ],
         ]);
 
--- a/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
+++ b/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
@@ -2,6 +2,7 @@
 
 namespace eLife\Search\Api\Elasticsearch;
 
+use DateTimeImmutable;
 use eLife\Search\Api\Query\QueryBuilder;
 use eLife\Search\Api\Query\QueryExecutor;
 
@@ -146,5 +147,18 @@ final class ElasticQueryBuilder implements QueryBuilder
         $exec->setQuery($this);
 
         return $exec;
+    }
+
+    public function betweenDates(DateTimeImmutable $fromDate, DateTimeImmutable $toDate): QueryBuilder
+    {
+        $this->query('range', [
+            'sortDate' => [
+                'gte' => $fromDate->format('Y/m/d'),
+                'lt' => $toDate->format('Y/m/d'),
+                'format' => 'yyyy/MM/dd||yyyy/MM/dd',
+            ],
+        ]);
+
+        return $this;
     }
 }
--- a/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
+++ b/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
@@ -159,7 +159,7 @@ final class ElasticQueryBuilder implements QueryBuilder
             $query['gte'] = $fromDate->format(self::ELASTIC_DATETIME_FORMAT);
         }
         if ($toDate) {
-            $query['lt'] = $toDate->format(self::ELASTIC_DATETIME_FORMAT);
+            $query['lte'] = $toDate->format(self::ELASTIC_DATETIME_FORMAT);
         }
         $this->query('range', [
             'sortDate' => $query,
--- a/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
+++ b/src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
@@ -11,6 +11,9 @@ final class ElasticQueryBuilder implements QueryBuilder
     private $order;
     private $exec;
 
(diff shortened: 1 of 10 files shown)


src/Search/Api/SearchController.php
```diff
@@ -54,7 +56,27 @@ public function indexAction(Request $request)
...
+        $fromDate = $request->query->get('fromDate');
+        $toDate = $request->query->get('toDate');
+        $toDateTime = null;
+        $fromDateTime = null;
+
+        if (
+            ($fromDate !== null && $toDate === null) ||
+            ($fromDate === null && $toDate !== null)
+        ) {
+            throw new BadRequestHttpException('You must provide both toDate and fromDate');
```
thewilkybarkid: Is this needed? The RAML doesn't define it like this.

src/Search/Api/SearchController.php
```diff
@@ -54,7 +56,27 @@ public function indexAction(Request $request)
...
+        }
+
+        if ($toDate && $fromDate) {
+            $toDateTime = DateTimeImmutable::createFromFormat('Y-m-d', $toDate);
+            $fromDateTime = DateTimeImmutable::createFromFormat('Y-m-d', $fromDate);
+            if ($toDateTime === false || $fromDateTime === false) {
```
thewilkybarkid: You should also check for things like `2017-02-29`.

thewilkybarkid: Also that `$fromDateTime <= $toDateTime`.

src/Search/Api/SearchController.php
```diff
@@ -54,7 +56,27 @@ public function indexAction(Request $request)
...
+
+        if (
+            ($fromDate !== null && $toDate === null) ||
+            ($fromDate === null && $toDate !== null)
+        ) {
+            throw new BadRequestHttpException('You must provide both toDate and fromDate');
+        }
+
+        if ($toDate && $fromDate) {
+            $toDateTime = DateTimeImmutable::createFromFormat('Y-m-d', $toDate);
```
thewilkybarkid: The time should be set to 23:59:59.

src/Search/Api/SearchController.php
```diff
@@ -54,7 +56,27 @@ public function indexAction(Request $request)
...
+            $fromDateTime = DateTimeImmutable::createFromFormat('Y-m-d', $fromDate);
```
thewilkybarkid: The time should be set to 00:00:00.

src/Search/Api/SearchController.php
thewilkybarkid: Make sure it's UTC too.

src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
```diff
@@ -147,4 +148,17 @@ public function getQuery(): QueryExecutor
 
         return $exec;
     }
+
+    public function betweenDates(DateTimeImmutable $fromDate, DateTimeImmutable $toDate): QueryBuilder
+    {
+        $this->query('range', [
+            'sortDate' => [
+                'gte' => $fromDate->format('Y/m/d'),
```
giorgiosironi: are we storing the date in this format?

stephenwf: On the way in it gets stored as a long:
`[ 2017-02-03T10:27:54Z ] -> [ elasticsearch ] -> [Long(1486117674)]`

so anything we query and on the way out again it gets formatted back

stephenwf:

Could you give this another look over, I've strengthened up the validation and made sure the times correlate. I don't think we need both the time and the inclusive date range, but since its all converted to the same numbers at the end of the line.. it won't make a difference

stephenwf:
Just finishing up installing goaws on our local build, then going to write a quick web test to cover this then we can merge

tests/src/Search/Web/ElasticTestCase.php
```diff
@@ -139,6 +139,7 @@ public function getJsonResponse()
     {
         $response = $this->getResponse();
         if (!$response->isOk()) {
+            var_dump($response);
```
giorgiosironi: It is usually preferable to put this into the `fail()` argument? `var_export()` supports returning instead of printing. By the way, isn't there a method for `$response` to give back a readable representation instead of the raw dump?

src/Search/Console.php
```diff
@@ -51,6 +54,17 @@
         ],
     ];
 
+    public function queueCreateCommand()
+    {
+        /** @var SqsClient $queue */
```
giorgiosironi: annotate/tag/describe this as development-only: queues should never be created manually on the real AWS

src/Search/Console.php
```diff
@@ -51,6 +54,17 @@
...
+        $queue = $this->app->get('aws.sqs');
+
+        $queue->createQueue([
```
giorgiosironi: method createQueue() on a queue? This is `$sqs`

tests/src/Search/Web/ElasticTestCase.php
stephenwf: Ah sorry forgot to remove that, I only used that for debugging. I've moved it into the `fail(...)`

src/Search/Api/Query/QueryBuilder.php
```diff
@@ -18,6 +20,8 @@ public function whereSubjects(array $subjects = []) : QueryBuilder;
 
     public function whereType(array $types = []) : QueryBuilder;
 
+    public function betweenDates(DateTimeImmutable $fromDate, DateTimeImmutable $toDate) : QueryBuilder;
```
thewilkybarkid: Aren't these nullable?

stephenwf: Good spot, fixed

PR #104, opened 2017-02-06
Fix start-date and end-date query string parameter names

thewilkybarkid:


tests/src/Search/Web/DateRangeTest.php
```diff
@@ -16,11 +16,11 @@ public function test_date_range_from_only()
         ]);
 
         $this->newClient();
-        $this->jsonRequest('GET', '/search', ['fromDate' => '2016-12-01']);
+        $this->jsonRequest('GET', '/search', ['start-date' => '2016-12-01']);
         $response = $this->getJsonResponse();
         $this->assertEquals($response->total, 2);
 
-        $this->jsonRequest('GET', '/search', ['fromDate' => '2016-11-01']);
+        $this->jsonRequest('GET', '/search', ['end-date' => '2016-11-01']);
```
stephenwf: This should be `start-date`

PR #107, opened 2017-02-09
ELPP-2093 Fixed subject and type query

stephenwf:
This should now return results in the correct format, without 500 errors and with all subjects and types intact. I'm using (a fairly accurate) workaround for the subject names pending the bus-sdk fixes. This allows this fix to be fast tracked (not requiring a reindexing) for demos and such.


stephenwf:
@giorgiosironi @thewilkybarkid Good to merge now?

thewilkybarkid:
Is `getSubjectName()` turning a subject ID (from the API) into name? If so, it won't be right (the ID won't have 'and' in it).

PR #110, opened 2017-03-01
Update SDK for reference change

thewilkybarkid:


giorgiosironi:
The failure was a salt highstate running at the same time, which shouldn't happen. Retrying.

PR #112, opened 2017-03-20
ELPP-2451 Added cache headers to search

stephenwf:


stephenwf:
Added a web test case, I will do the same with Recommendations when it gets a web test framework installed.

thewilkybarkid:
Could you also add one that checks that the `Cache-Control` header is correct?

stephenwf:
added!

tests/src/Search/Web/CacheWebTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->addDocumentsToElasticSearch([
+            $this->getArticleFixture(0),
+            $this->getArticleFixture(1),
+        ]);
+        $this->newClient();
+        $this->jsonRequest('GET', '/search');
+        $response = $this->getResponse();
+        $this->assertEquals(200, $response->getStatusCode());
+        $eTag = $response->headers->get('ETag');
+        $this->assertEquals('max-age=300, public', $response->headers->get('Cache-Control'));
```
thewilkybarkid: Shouldn't this fail? `stale-while-revalidate` and `stale-if-error` should be there.

thewilkybarkid:
Should check the `Accept` header too.

tests/src/Search/Web/CacheWebTest.php
stephenwf: That is my bad, added those to the wrong header!

PR #113, opened 2017-03-22
Removed HTTP cache for search

stephenwf:


thewilkybarkid:
Can `doctrine/cache` go too?

stephenwf:
We are using doctrine cache for the rest of the app, so for annotations and metadata.

composer.lock
```diff
@@ -4,7 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "0ceb266d9819ca0d12f1388267bc7375",
+    "hash": "de7fb5224e4a1eb6ae01d38acb37c023",
+    "content-hash": "d171c3ed89ae5dc94390939ef9f1514e",
```
thewilkybarkid: A lot of these changes are due to using an old version of Composer. Redo this after a `composer self-update`.

stephenwf: No more puli, ah so we can use version 1.4.1

thewilkybarkid: 🎉

stephenwf: Running a composer update while I'm at it

PR #118, opened 2017-04-12
ELPP-2569 Change images to IIIF

thewilkybarkid:
There's no transition period here, so needs to be deployed at the same time as the Journal update.


giorgiosironi:
This must be merged now, because the project is not able to build with the new api-dummy.

PR #122, opened 2017-05-18
Manual atomic reindex commands

giorgiosironi:


src/Search/Api/Elasticsearch/ElasticsearchClient.php
```diff
@@ -18,6 +18,12 @@ public function __construct(Client $connection, string $index, bool $forceSync =
         $this->forceSync = $forceSync;
     }
 
+    public function setIndex($indexName){
+
+        $this->index = $indexName;
```
giorgiosironi: would rename this in `$this->defaultIndex` as it seems this class will be able to work with more than one at the same time, when `$index` is passed as a parameter to its methods.

src/Search/Api/Elasticsearch/ElasticsearchClient.php
```diff
@@ -102,4 +121,31 @@ public function getDocumentById($type, $id)
...
+                ),
+                'dest' => array(
+                    'index' => $destination
+                )
+            )
+        );
+
+        return $this->connection->reindex($params);
+
+    }
```
giorgiosironi: all these primitive seem useful to manage multiple indexes :+1:

src/Search/Console.php
```diff
@@ -119,6 +122,75 @@ private function enqueue($type, $id)
...
+
+        }
+
+        // Kill all existing Gearnman workers
+        // Giorgio perhaps some linux command
+
+        // start some new ones using our new index
+        // Populate exiting items
+        //subsribe to SQS queue
+        exec('./bin/console gearman:worker  --index="elife_search_tmp" >> /tmp/gearman-worker.log 2>&1 &');
```
giorgiosironi: I think this section has to be managed outside of this script. We could make the assumption that nothing is running here, and just populate the new index without touching the existing one. From the deployment, we will be able to:
1. stop all workers
1. call `indexRebuildCommand()`
1. switch the index
1. start all workers again

During this process, the application will always be able to serve reads:
- the old, stale index
- the new, stale index as it catches up (after the switch)
- the new, up-to-date index (a few seconds/minutes after the switch)

So I think this command should be limited to: build a new index with a custom name.
Then we have the problem of switching, but I would separate that problem into another command as we have to decide:
- how to store the index name (some config file)
- how to switch (rewrite the file?)

src/Search/Api/Elasticsearch/ElasticsearchClient.php
```diff
@@ -28,9 +34,22 @@ public function deleteIndexByName($index)
...
+
+        $params = [
+            'index' => $indexName
+
+        ];
+        return $this->connection->indices()->create($params);
+
+    }
+
+    public function deleteIndex($indexName ='')
```
stephenwf: Would be good to get some type hints on the parameter, and a default of `null`

src/Search/Api/Elasticsearch/ElasticsearchClient.php
```diff
@@ -28,9 +34,22 @@ public function deleteIndexByName($index)
...
     {
-        return $this->deleteIndexByName($this->index);
+        if (empty($indexName)){
+            $indexName = $this->index;
+        }
+        return $this->deleteIndexByName($indexName);
```
stephenwf: If the default is null, this could be simplified to `$indexName ?? $this->index` (applied to other methods too)

src/Search/Api/Elasticsearch/ElasticsearchClient.php
```diff
@@ -102,4 +121,31 @@ public function getDocumentById($type, $id)
...
+            'index'=>$index,
+        ];
+
+        return $this->connection->count($params);
+
+    }
+
+    public function moveIndex($source, $destination){
+
+        $params = array(
```
stephenwf: For consistency sake, can these be changed to use the short hand array syntax

src/Search/Console.php
```diff
@@ -119,6 +122,75 @@ private function enqueue($type, $id)
...
+    public function indexRebuildCommand(){
+
+        $newIndexName = 'elife_search_tmp';
+        $client = $this->getElasticClient();
+
+        // Delete the existing 'elife_search_tmp' index if we have one
+        echo "checking if an old 'elife_search_tmp' index exists \n";
+        $response = $client->deleteIndex($newIndexName);
+
+        if ($response['payload'] instanceof \eLife\Search\Api\Elasticsearch\Response\SuccessResponse){
```
stephenwf: Again, for consistency and readability, can you import these namespaces

src/Search/Console.php
```diff
@@ -119,6 +122,75 @@ private function enqueue($type, $id)
...
+
+            echo "Found an old 'elife_search_tmp' index, Deleted it \n";
```
stephenwf: These should go through the:
```
$this->logger->error();
// or
$this->logger->debug();
```

src/Search/Api/Elasticsearch/ElasticsearchClient.php
```diff
@@ -102,4 +121,31 @@ public function getDocumentById($type, $id)
 
         return $this->connection->get($params)['payload'] ?? null;
     }
+
+    public function count($index){
```
stephenwf: Needs a run through the PHP-CS-Fixer

src/Search/Console.php
```diff
@@ -119,6 +122,75 @@ private function enqueue($type, $id)
...
+        $response = $client->createIndex($newIndexName);
+
+        if ($response['payload'] instanceof \eLife\Search\Api\Elasticsearch\Response\SuccessResponse){
+
+            echo "Successfully created the new index \n";
+
+        }else {
+
+            echo "Could not create the new index ... exiting \n";
+            die();
```
stephenwf: Instead of dying here, you should be able to `return 1;` where `1` is the exit code

src/Search/Console.php
```diff
@@ -119,6 +122,75 @@ private function enqueue($type, $id)
...
+            $client->deleteIndex('elife_search');
+            $client->createIndex('elife_search');
+            $client->moveIndex('elife_search_tmp','elife_search');
+
+            // We should kill the workers we started that goto elife_search_tmp as it no longer exists
+
+        }else{
+
+            echo "Error: New index count wasn't greater than old index. Restart the gearman workers on the old index to bring it back form being stale \n";
+            die();
```
stephenwf: `return 1;`

giorgiosironi:
Yes, this solution is only tested for reindexing where the index schema does not change. We can then build on top a backward compatibility layer at the first index schema change (will depend on how much it changes).

PR #129, opened 2017-06-23
Automate reindexing

giorgiosironi:
The script:

1. stops gearman workers
2. restarts custom gearman workers writing to the new index (actually 1)
3. imports all in the new index
4. waits for queues to be empty

However, before continuing with switching the index I want to make the workers from step 2 more reliable. From experience in prod, they are continuously being stopped and restarted when they run out of memory.

Therefore I would like to change the configuration of the index in something like:
```
// index.json
{
    "write": "elife_search_48",
    "read": "elife_search_47"
}
```
which was probably one of the original designs. This would make the workers from 2 live under Upstart like the normal ones, we just need to change the configuration and restart them.


thewilkybarkid:
Rather than separate reading and writing, can't the workers just be restarted after the new index is live?

giorgiosironi:
In phase 3 we need (some) workers to be alive to write to index B, but simultaneously API requests should be served by index A; the constraint is we need workers with a modified configuration to create the new index B while the import is performed.

thewilkybarkid:
Ah yeah, they go through the same workers.

thewilkybarkid:
👍  to the `index.json` proposal

giorgiosironi:
Should be ready for review, need to keep some backward compatibility with `index.txt` in for now. It's not executed automatically but after tests in end2end and prod it can be executed in the formula, according to some conditions to be checked like a version file increment.

PR #131, opened 2017-06-26
Reindexing on demand

giorgiosironi:
Modify the index.import file with a new data to trigger a reindex in all the environments that the code is deployed too. The formula will call `bin/reindex-on-demand`.

The (approximate) date of the last import is stored in `index.json` after being retrieved from `index.import`. When the committed file changes, a reindexing and an update of `index.json` is triggered.

Interaction with `index.json` is performed through `bin/console` commands.


Jenkinsfile
```diff
@@ -24,6 +24,7 @@ elifePipeline {
                         builderSmokeTests 'search--end2end', '/srv/search'
                         builderCmd 'search--end2end', 'cd /srv/search; php bin/console queue:import all --env=end2end'
                         builderCmd 'search--end2end', 'cd /srv/search; ./bin/wait-for-empty-queues end2end'
+                        builderCmd 'search--end2end', "cd /srv/search; ./bin/reindex end2end elife_search_${env.BUILD_NUMBER}"
```
thewilkybarkid: These should replace the above two commands shouldn't it?

giorgiosironi: This is a testing environment, so we were attempting to test both an import from scratch and a reindexing. But the import from scratch is not cleaning the db first, so it should do that. Then we could change the order: 1. update the existing index, then 2. test also the import from scratch.

PR #132, opened 2017-07-27
Build that updates the SDK and some of its dependencies

giorgiosironi:
Give a man a :fish: ...

### DIFF
--- a/Jenkinsfile.update-api-sdk-php
+++ b/Jenkinsfile.update-api-sdk-php
@@ -1,0 +1,15 @@
+elifeUpdatePipeline(
+    { commit ->
+        lock('search--ci') {
+            builderDeployRevision 'search--ci', commit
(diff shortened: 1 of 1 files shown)


Jenkinsfile.update-api-sdk-php
```diff
@@ -0,0 +1,15 @@
+elifeUpdatePipeline(
+    { commit ->
+        lock('search--ci') {
+            builderDeployRevision 'search--ci', commit
+            builderCmd "search--ci", "composer update elife/api elife/api-client elife/api-sdk --no-interaction", "/srv/search"
```
giorgiosironi: @thewilkybarkid is the list of dependencies to update right? Don't want locks on the other ones keeping api-sdk from being updated

thewilkybarkid: Might be best to go `elife/api elife/api-sdk --with-dependencies --no-suggest` (and treat `elife/api-client` as a dependency of `elife/api-sdk`).

PR #138, opened 2017-08-04
ELPP-2601 Interview image

thewilkybarkid:


thewilkybarkid:
Not sure if this needs a reindex or not, there won't be data at that point.

api-dummy.sha1
```diff
@@ -1 +1 @@
-49b90e5f6fec86bf6785e198e9ca65d46a5937c1
+420e0cc99b20c98eb6f415bcf78815c51508dd6f
```
giorgiosironi: should we do a similar version of https://ci--alfred.elifesciences.org/job/dependencies-search-update-api-sdk-php/ for api-dummy and run it weekly?

tests/src/Search/Web/ExampleWebTest.php
```diff
@@ -33,7 +33,7 @@ public function testElasticSearchIndex()
                 ],
```
giorgiosironi: tests/src/Search/Api/Elasticsearch/InterviewElasticsearchTest.php should need an update? It uses the minimum and complete JSONs so there should be one with an actual image being returned and checked?

thewilkybarkid: Yeah, was going to talk to you about testing. Wondering if `testSerializationSmokeTest()` should be expanded to test that it produces the same results as the SDK serialization? (Which raises the bigger problem that the SDK serialization should be used instead.)

giorgiosironi: Don't know the effort required but it would save lines of code to maintain on the test side. I am not sure if the duplicate serialization (these yaml configuration) is necessary for correct indexing of the data.

PR #144, opened 2017-08-23
ELPP-2908 Change for parameter to be AND rather than OR

thewilkybarkid:
Also adds fuzzy matching.

### DIFF
--- a/tests/src/Search/Web/ForTest.php
+++ b/tests/src/Search/Web/ForTest.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace tests\eLife\Search\Web;
+
+/**
+ * @group web
+ */
+final class ForTest extends ElasticTestCase
+{
+    public function test_all_for_words_match()
+    {
+        $this->addDocumentsToElasticSearch([
+            $this->getArticleFixture(0),
+            $this->getArticleFixture(1),
+            $this->getArticleFixture(2),
+            $this->getArticleFixture(3),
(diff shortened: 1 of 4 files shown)


thewilkybarkid:
Not sure if these need to be applied the boostings.

tests/src/Search/Web/ForTest.php
```diff
@@ -0,0 +1,43 @@
...
+            $this->getArticleFixture(3),
+        ]);
+
+        $this->newClient();
+        $this->jsonRequest('GET', '/search', ['for' => 'Robert Smiht']);
+        $response = $this->getJsonResponse();
+        $this->assertEquals($response->total, 2);
+        $this->assertEquals($response->items[0]->id, '15278');
+        $this->assertEquals($response->items[1]->id, '15276');
+    }
```
giorgiosironi: would add `Robert $inventedWord` and check `total` is `0`

thewilkybarkid: 👍

thewilkybarkid:
Looking at it, we might want to use [simple_query_string](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-simple-query-string-query.html), as that allows for phrases (eg `foo "bar baz"`).

thewilkybarkid:
Hmm, not sure how that works with boosting. Will leave for now.

PR #147, opened 2017-09-06
Fix author response blocks

thewilkybarkid:
Relates to PI-498.

Looks like this will need to be extracted and used in lots of places, but the author response is currently causing a problem.


scottmelville:
Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #155, opened 2017-10-11
ELPP-3153 Normalise all cases where there can be a paragraph and quote block

thewilkybarkid:
I've taken the quick option and just reapply the existing logic to all cases where there can be a paragraph and quote block, which should allow all content to be indexed. I've dropped storing the original JSON as I don't believe it's used.

I want to make some larger changes that would solve this problem for good, but this should  be enough for now.


thewilkybarkid:
Error looks to be: `{\"error\":{\"root_cause\":[{\"type\":\"mapper_parsing_exception\",\"reason\":\"object mapping for [content] tried to parse field [null] as object, but found a concrete value\"}],\"type\":\"mapper_parsing_exception\",\"reason\":\"object mapping for [content] tried to parse field [null] as object, but found a concrete value\"},\"status\":400}` on a blog article.

giorgiosironi:
All the data come from the api-dummy, are we able to log the id of that item?

thewilkybarkid:
Not logged apparently. Going to try and run locally.

thewilkybarkid:
Failed to recreate locally.

thewilkybarkid:
Since this is changing types inside ES I think it needs an immediate reindex.

thewilkybarkid:
I ended up just turning blocks into a single string. We're not doing anything clever with weighting of specific things (eg figure titles) yet, so should be ok.

src/Search/Workflow/BlogArticleWorkflow.php
```diff
@@ -86,6 +87,8 @@ public function index(BlogArticle $blogArticle) : array
         $this->logger->debug('BlogArticle<'.$blogArticle->getId().'> Indexing '.$blogArticle->getTitle());
         // Normalized fields.
         $blogArticleObject = json_decode($this->serialize($blogArticle));
+        $blogArticleObject->body = $this->flattenBlocks($blogArticleObject->content ?? []);
+        unset($blogArticleObject->content);
```
thewilkybarkid: Conflicts with the content of a collection otherwise.

src/Search/Workflow/Blocks.php
```diff
@@ -0,0 +1,28 @@
...
+        return implode(' ', array_filter([
+            $block->id ?? null,
+            $block->label ?? null,
+            $block->title ?? null,
+            $block->alt ?? null,
+            $block->text ?? null,
+            $block->caption->text ?? null,
+            $block->question ?? null,
+            $this->flattenBlocks($block->answer ?? []),
+            $this->flattenBlocks($block->content ?? []),
```
thewilkybarkid: This isn't every single property that can exist on a block, but I think it covers the important ones.

thewilkybarkid:
Can we try this on a different environment to make sure it can successfully index everything?

src/Search/Api/Elasticsearch/Command/resources/labs-post.yaml
```diff
@@ -1,35 +1,5 @@
 labs-post:
   properties:
-    content:
```
giorgiosironi: does this mean the same information goes into another field defined here or simply that we don't define these fields explicitly?

giorgiosironi:
`end2end` always reindexes its content: https://github.com/elifesciences/search/blob/develop/Jenkinsfile#L25 However it's not much content as it's only that produced by the tests at the moment.
If you want to use real production data, perhaps it's easier to substitute the api-dummy configuration with the real API in the Vagrant VM. Sort of changing `api_url` in `config/dev.php`?

thewilkybarkid:
I indexed all items locally, increased the number of items from 5015 to 5024 (some things may have been published in-between, but one known to have been missing is now there).

thewilkybarkid:
Also nothing appeared in the error log.

PR #158, opened 2017-10-17
Index list items

thewilkybarkid:


thewilkybarkid:
No re-index as previous reindex hasn't been triggered yet.

src/Search/Workflow/Blocks.php
```diff
@@ -23,6 +23,18 @@
             $block->question ?? null,
             $this->flattenBlocks($block->answer ?? []),
             $this->flattenBlocks($block->content ?? []),
+            $this->flattenItems($block->items ?? []),
```
giorgiosironi: ok, `"prefix": "roman-lower"` is excluded

giorgiosironi: there can be HTML in `"text"` such as `"<i>P. khe59svzdsh80x9qhhhzon6m6fqqo2</i>"`. Maybe not a problem

thewilkybarkid: Can be HTML already in lots of places. We could probably strip it.

PR #167, opened 2017-11-09
Add support for a config.php file

giorgiosironi:
Still allows to fall back to a environment-specific file.

Removes the `local` environment as it's a mess to setup and maintain, if we want something different from a VM we should look into a set of containers.


giorgiosironi:
Could do much more, but don't wanna introduce conflicts.

src/Search/Console.php
```diff
@@ -210,7 +210,7 @@ public function __construct(Application $console, Kernel $app)
         // Some annotations
         Register::registerLoader();
 
-        $this->console->getDefinition()->addOption(new InputOption('--env', '-e', InputOption::VALUE_REQUIRED, 'The Environment name.', 'dev'));
+        $this->console->getDefinition()->addOption(new InputOption('--env', '-e', InputOption::VALUE_OPTIONAL, 'The Environment name. Deprecated and not used', 'dev'));
```
thewilkybarkid: Guess setting it causes a problem?

bin/console
```diff
@@ -8,18 +8,10 @@ use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Output\ConsoleOutput;
 
-require_once __DIR__.'/../vendor/autoload.php';
-
-umask(002);
+require_once __DIR__.'/../bootstrap.php';
```
thewilkybarkid: Move the bootstrap into `src`?

bootstrap.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+require_once __DIR__.'/vendor/autoload.php';
+umask(0002);
+$defaultFile = __DIR__.'/config/config.php';
+$legacyFile = __DIR__.'/config/'.getenv('ENVIRONMENT_NAME').'.php';
+if (file_exists($defaultFile)) {
+    $config = include $defaultFile;
```
thewilkybarkid: Take a look at https://github.com/elifesciences/recommendations/pull/100 where I'm working on a slightly different approach.

src/Search/Console.php
giorgiosironi: Could not cause a problem, there is is some wrangling of `--env` in `bin/console`, so put this as a precaution to remove it down the line.

bin/console
giorgiosironi: Can do

bootstrap.php
giorgiosironi: Should be easy to add the environment-specific files, merge them, move the main one up a folder

giorgiosironi:
Done, tests (for now) pass.

PR #168, opened 2017-11-10
ELPP-3180 Implement arbitrary store and load

giorgiosironi:
The goal is to use the same object to store `IndexMetadata`, which is currently stored on the filesystem.

Creates a separate index, which we can set up to perform no processing and store arbitrary data.

The application will eventually load metadata from there to know which index should be used for reading/writing, plus other ancillary information store in the same document.

```
[17:04:25][elife@dev--search:/srv/search]$ php bin/console arbitrary:store my_id '{"index":"value7"}' --env=dev
[17:04:29][elife@dev--search:/srv/search]$ php bin/console arbitrary:load my_id --env=dev
array (
    '_index' => 'arbitrary',
    '_type' => 'json_object',
    '_id' => 'my_id',
    '_version' => 8,
    'found' => true,
    '_source' =>
    array (
        'index' => 'value7',
    ),
)
```

### DIFF
--- a/bin/ci-import
+++ b/bin/ci-import
@@ -2,6 +2,8 @@
 . bin/ci-process-bootstrap
 
 echo '{"write":"elife_search", "read":"elife_search"}' > index.json
+# FUTURE:
+# php bin/console keyvalue:store index-metadata '{"write":"elife_search", "read":"elife_search"}' --env="$environment"
 
 ./bin/console queue:clean --env="$environment"
 # TODO: poll for queue:count == 0, with timeout
--- a/src/Search/ArbitraryDataRepository.php
+++ b/src/Search/ArbitraryDataRepository.php
@@ -42,4 +42,13 @@ final class ArbitraryDataRepository
             self::INDEX_NAME
         );
     }
+
+    public function load(string $key) : array
+    {
+        return $this->client->getPlainDocumentById(
+            self::DOCUMENT_TYPE,
+            $key,
+            self::INDEX_NAME
+        );
+    }
 }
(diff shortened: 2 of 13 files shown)


src/Search/ArbitraryDataRepository.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace eLife\Search;
+
+use eLife\Search\Api\Elasticsearch\ElasticsearchClient;
+
+// TODO: rename to ElasticsearchArbitraryDataRepository
+// TODO: Extract Interface with the current name
+final class ArbitraryDataRepository
```
thewilkybarkid: `KeyValueStore`?

bin/ci-import
```diff
@@ -2,6 +2,8 @@
 . bin/ci-process-bootstrap
 
 echo '{"write":"elife_search", "read":"elife_search"}' > index.json
+# FUTURE:
+# php bin/console keyvalue:store index-metadata '{"write":"elife_search", "read":"elife_search"}' --env="$environment"
```
thewilkybarkid: I spy a new library.

giorgiosironi: Any project where it would be useful? I thought it would be a debug tool, mostly; targeting different backends than ElasticSearch?

thewilkybarkid: Journal's going to need a KV store soon (probably with a DB?). There are a couple of existing KV libraries, but they don't seem to be maintained.

giorgiosironi:
This should be ready for review and merge for the first step, which is to deploy and then add the `keyvalue:setup` command to the formula. After that, we can port the `end2end`/`prod` data (perhaps manually) and switch the calls.

bin/ci-import
giorgiosironi: Ok, will extract after the search ordeal, when stable, with only the existing backend. Adding to ELPP-3180 checklist.

PR #170, opened 2017-11-13
Reading from key value store

giorgiosironi:
Got to this point before the build started misteriously failing.

PR #171, opened 2017-11-14
Fixed false positive on 500 during expected results check

giorgiosironi:


giorgiosironi:
Extracted from #170 as an urgent bug fix.

thewilkybarkid:
What was happening?

giorgiosironi:
500 response would produce an unparseable JSON, which meant the bash `if` would do `! -eq 0` which is sort-of true.

PR #172, opened 2017-11-15
Forcing ENVIRONMENT_NAME from cli argument

giorgiosironi:
Upstart does not provide environment variables, so this is the only way
to provide the environment at the moment.


giorgiosironi:
Will merge and deploy tomorrow.

PR #174, opened 2017-11-16
Two different clients

giorgiosironi:
The current Client class for ElasticSearch was overloaded with:
- methods that only work when configured to map responses onto objects
- methods that only work when configured to return plain response bodies

We differentiate them into two classes, remove dead code, and make the dependencies of commands towards them more precise.


src/Search/Api/Elasticsearch/MappedElasticsearchClient.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $con = $this->libraryClient->index($params)['payload'] ?? null;
+        if ($flush || $this->forceSync) {
+            $this->libraryClient->indices()->refresh(['index' => $this->index]);
+        }
+
+        return $con;
+    }
+
+    // mapped, if used
```
thewilkybarkid: Guess these can go.

giorgiosironi: I hoped so, but https://github.com/elifesciences/search/blob/develop/src/Search/Workflow/ResearchArticleWorkflow.php#L181

PR #181, opened 2017-12-08
Date filters exclude boundaries

giorgiosironi:
Excluding because the variables were mixed up and were setting 23:59:59 for the start date and 00:00:00 for the end date.

### DIFF
--- a/src/Search/Workflow/ResearchArticleWorkflow.php
+++ b/src/Search/Workflow/ResearchArticleWorkflow.php
@@ -132,7 +132,7 @@ final class ResearchArticleWorkflow implements Workflow
             'value' => json_encode($articleObject->dataSets ?? '[]'),
         ];
 
-        $sortDate = $article->getStatusDate() ? $article->getStatusDate() : $article->getPublishedDate();
+        $sortDate = $article->getStatusDate();
         $this->addSortDate($articleObject, $sortDate);
 
         $this->logger->debug('Article<'.$article->getId().'> Detected type '.($article->getType() ?? 'research-article'));
--- a/tests/src/Search/Web/DateRangeTest.php
+++ b/tests/src/Search/Web/DateRangeTest.php
@@ -23,6 +23,11 @@ class DateRangeTest extends ElasticTestCase
         $this->jsonRequest('GET', '/search', ['start-date' => '2016-11-01']);
         $response = $this->getJsonResponse();
         $this->assertEquals($response->total, 3);
+
+        // boundary is included
+        $this->jsonRequest('GET', '/search', ['start-date' => '2016-12-19']);
+        $response = $this->getJsonResponse();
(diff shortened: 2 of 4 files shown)


src/Search/Workflow/ResearchArticleWorkflow.php
```diff
@@ -132,7 +132,7 @@ public function index(ArticleVersion $article) : array
             'value' => json_encode($articleObject->dataSets ?? '[]'),
         ];
 
-        $sortDate = $article->getStatusDate() ? $article->getStatusDate() : $article->getPublishedDate();
+        $sortDate = $article->getStatusDate();
```
thewilkybarkid: Makes sense to remove. If there's no status date, there won't be a published date either.

tests/src/Search/Web/DateRangeTest.php
```diff
@@ -34,13 +39,18 @@ public function test_date_range_end_only()
...
 
         $this->jsonRequest('GET', '/search', ['end-date' => '2016-10-01']);
         $response = $this->getJsonResponse();
-        $this->assertEquals($response->total, 0);
+        $this->assertEquals(0, $response->total);
+
+        // boundary
+        $this->jsonRequest('GET', '/search', ['end-date' => '2016-12-05']);
+        $response = $this->getJsonResponse();
+        $this->assertEquals(1, $response->total, 'The date upper boundary is being excluded');
```
thewilkybarkid: Could we list the IDs on the included results? Would make reading the tests easier.

tests/src/Search/Web/DateRangeTest.php
```diff
@@ -16,13 +18,18 @@ public function test_date_range_start_only()
...
         $response = $this->getJsonResponse();
-        $this->assertEquals($response->total, 2);
+        $this->assertIds(['15275', '15276'], $response);
 
         $this->jsonRequest('GET', '/search', ['start-date' => '2016-11-01']);
         $response = $this->getJsonResponse();
-        $this->assertEquals($response->total, 3);
+        $this->assertIds(['15275', '15276', '19662'], $response);
+
+        // boundary is included
```
thewilkybarkid: The previous assertion could be for `2016-12-05` instead?

giorgiosironi: It's still two different cases: the previous one is a filter that allows all results, while the following is a filter that falls on one of the item dates. Not worth it to go property-based though...

thewilkybarkid: Ok; 19662 has a time (ie it's not `00:00:00`), so either use that here or change 15276 to also have a time.

giorgiosironi: The API only allows days, not times, so changing 15276

PR #182, opened 2017-12-11
Cache field is populated, but not used

giorgiosironi:
Found while understanding what search is caching.

### DIFF
--- a/src/Search/Api/SearchController.php
+++ b/src/Search/Api/SearchController.php
@@ -4,7 +4,6 @@ namespace eLife\Search\Api;
 
 use DateTimeImmutable;
 use DateTimeZone;
-use Doctrine\Common\Cache\Cache;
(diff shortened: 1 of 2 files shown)


src/Search/Api/SearchController.php
```diff
@@ -4,7 +4,6 @@
 
 use DateTimeImmutable;
 use DateTimeZone;
-use Doctrine\Common\Cache\Cache;
```
thewilkybarkid: `doctrine/cache` can probably be removed from `composer.json`?

giorgiosironi: It seems used to cache annotations: https://github.com/elifesciences/search/blob/800e57dc392694ca7763e6174963552f70a5df58/src/Search/Kernel.php#L166-L180

PR #183, opened 2017-12-15
Use csa/guzzle-cache-middleware library

thewilkybarkid:
Has been extracted from the Symfony bundle so can reuse easily.


tests/src/Search/HttpClient.php
```diff
@@ -3,14 +3,13 @@
...
 use ComposerLocator;
-use Csa\Bundle\GuzzleBundle\GuzzleHttp\Middleware\MockMiddleware;
+use Csa\GuzzleHttp\Middleware\Cache\MockMiddleware;
 use eLife\ApiClient\HttpClient\Guzzle6HttpClient;
 use eLife\ApiValidator\MessageValidator\JsonMessageValidator;
 use eLife\ApiValidator\SchemaFinder\PathBasedSchemaFinder;
 use GuzzleHttp\Client;
 use GuzzleHttp\HandlerStack;
 use JsonSchema\Validator;
-use test\eLife\ApiSdk\ValidatingStorageAdapter;
```
thewilkybarkid: I'd forgotten that this uses a library's test code. :shudders:

tests/src/Search/ValidatingStorageAdapter.php
```diff
@@ -0,0 +1,51 @@
...
+use eLife\ApiValidator\Exception\InvalidMessage;
+use eLife\ApiValidator\MessageValidator;
+use Psr\Http\Message\MessageInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use RuntimeException;
+use function GuzzleHttp\json_decode;
+use function GuzzleHttp\json_encode;
+
+final class ValidatingStorageAdapter implements StorageAdapterInterface
```
thewilkybarkid: Will move to `eLife\ApiValidator`, since a few other projects use it.

tests/src/Search/HttpClient.php
giorgiosironi: removing the `use` without removing https://github.com/elifesciences/search/pull/183/files#diff-2a45a15b445df51cb0cb7005d3bf2f6cL30 ?

thewilkybarkid: Check the namespace.

giorgiosironi: ah, class "inlined" in the project

PR #191, opened 2018-02-06
ELPP-3433 Instrument to measure performance

giorgiosironi:
Experimenting with Stopwatch shows anything that is not Elasticsearch calls executes in <1.0ms locally.

Logs produced:
```
{"message":"Request Success:","context":{"method":"GET","uri":"http://localhost:9200/key_value_store/json-object/index-metadata","headers":{"host":["localhost:9200"]},"HTTP code":200,"duration":0.005583},"level":200,"level_name":"INFO","channel":"search","datetime":{"date":"2018-02-06 16:35:16.910163","timezone_type":3,"timezone":"UTC"},"extra":{"process_id":1238}}
{"message":"Request Success:","context":{"method":"GET","uri":"http://localhost:9200/key_value_store/json-object/index-metadata","headers":{"host":["localhost:9200"]},"HTTP code":200,"duration":0.001795},"level":200,"level_name":"INFO","channel":"search","datetime":{"date":"2018-02-06 16:35:16.912400","timezone_type":3,"timezone":"UTC"},"extra":{"process_id":1238}}
{"message":"Request Success:","context":{"method":"GET","uri":"http://localhost:9200/elife_search_20180129132425/_search?from=0&size=10","headers":{"host":["localhost:9200"]},"HTTP code":200,"duration":0.042386},"level":200,"level_name":"INFO","channel":"search","datetime":{"date":"2018-02-06 16:35:16.961486","timezone_type":3,"timezone":"UTC"},"extra":{"process_id":1238}}
```

Timeouts are overridable through configuration, but have sane defaults in one place.

The other operations may warrant a timeout too, but they seem less dangerous so I'd like to validate this first.


src/Search/Api/SearchController.php
```diff
@@ -134,6 +136,8 @@ public function indexAction(Request $request, Accept $type)
                 break;
         }
 
+        $query = $query->setClientOptions($this->clientOptions);
```
thewilkybarkid: Shouldn't be in the controller? Looking at the architecture, set in... `ElasticQueryExecutor`? (Which appears to be stateful. 😐) Though maybe `MappedElasticsearchClient` should actually take the query object and set it there instead.

src/Search/Kernel.php
```diff
@@ -86,6 +86,10 @@ public function __construct($config = [])
             'elastic_servers' => ['http://localhost:9200'],
             'elastic_logging' => false,
             'elastic_force_sync' => false,
+            'elastic_search_client_options' => [
+                'timeout' => 0.9,
+                'connect_timeout' => 0.5,
```
thewilkybarkid: 👍

giorgiosironi:
In the mean time, removed the middleman `ElasticQueryExecutor`, which didn't have any responsibility.

PR #193, opened 2018-02-13
ELPP-3447 504 on timeout from ElasticSearch

giorgiosironi:
Not using `api-problem` yet, but it can be added later if it fits.

### DIFF
--- a/src/Search/Api/SearchController.php
+++ b/src/Search/Api/SearchController.php
@@ -4,6 +4,7 @@ namespace eLife\Search\Api;
 
 use DateTimeImmutable;
 use DateTimeZone;
+use Elasticsearch\Common\Exceptions\NoNodesAvailableException;
 use eLife\ApiSdk\Model\Subject;
 use eLife\Search\Api\Elasticsearch\ElasticQueryBuilder;
 use eLife\Search\Api\Elasticsearch\MappedElasticsearchClient;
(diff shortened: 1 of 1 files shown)


src/Search/Api/SearchController.php
```diff
@@ -134,7 +136,11 @@ public function indexAction(Request $request, Accept $type)
                 break;
         }
 
-        $data = $this->client->searchDocuments($query->getRawQuery());
+        try {
+            $data = $this->client->searchDocuments($query->getRawQuery());
+        } catch (NoNodesAvailableException $e) {
```
thewilkybarkid: Does this wrap timeout errors from cURL?

giorgiosironi: It is what comes out from the black box ElasticSearch SDK (see https://elifesciences.atlassian.net/browse/ELPP-3447?focusedCommentId=30254&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-30254) as it doesn't expose the underlying requests and implements some (unused so far) retry logic.

thewilkybarkid: We can handle things as we see them. Was trying to understand how it works a bit (looks to unfortunately use RingPHP+cURL).

giorgiosironi: Uhm I wonder which of the twelve Guzzle major versions that corresponds to

thewilkybarkid: Part of the short-lived Guzzle 5.

giorgiosironi:
Got this also in https://alfred.elifesciences.org/job/test-journal-cms/428/consoleFull

PR #194, opened 2018-02-15
Reindex continuumtest

giorgiosironi:
Copying from what we do in `prod` (`Jenkinsfile.prod`). Not sure this is a solution as the reindex itself needs to work well first.


giorgiosironi:
Outdated.

PR #196, opened 2018-03-09
ELPP-3490 Increase maximum sizes of aggregated terms

giorgiosironi:


src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
```diff
@@ -16,14 +16,17 @@
     const DATE_DEFAULT = 'sortDate';
     const DATE_PUBLISHED = 'published';
 
+    const MAXIMUM_TYPES = 100;
```
thewilkybarkid: `MAXIMUM_SUBJECTS`?

src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
```diff
@@ -16,14 +16,17 @@
...
+
     private $dateType;
 
     public function __construct(string $index)
     {
         $this->query['index'] = $index;
         $this->query['body']['aggregations']['type_agg']['terms'] = [
             'field' => '_type',
-            'size' => 18,
+            'size' => self::MAXIMUM_TYPES,
```
thewilkybarkid: This is a constant number, so ok to leave as 18.

thewilkybarkid: Though I count it as 17!

giorgiosironi: I'll leave the constant and put it to 18

giorgiosironi: `curl -v https://prod--gateway.elifesciences.org/search | jq '.types | length'` is 17 in fact.

giorgiosironi: It seems like the kind of things that will break with no one noticing

giorgiosironi:
Let's see what this does.

PR #199, opened 2018-03-12
ELPP-3501 Add rate limit to high pages

giorgiosironi:
Should increment the counter by one.

Will be used by https://github.com/elifesciences/builder-private/pull/154


src/Search/Kernel.php
```diff
@@ -82,6 +82,7 @@ public function __construct($config = [])
             'api_url' => '',
             'api_requests_batch' => 10,
             'ttl' => 300,
+            'rate_limit_minimum_page' => 2,
```
giorgiosironi: will now configure this depending on environment

PR #202, opened 2018-03-14
ELPP-3163 Run proofreader 0.2 for forward compatibility

giorgiosironi:


src/Search/Annotation/GearmanTaskDriver.php
```diff
@@ -149,7 +149,7 @@ public function work()
             try {
                 $this->logger->debug('gearman:worker: Loop start, listening for jobs');
                 $result = $this->worker->work();
-                if ($this->worker->returnCode() == GEARMAN_TIMEOUT) {
+                if (GEARMAN_TIMEOUT == $this->worker->returnCode()) {
```
thewilkybarkid: Should be strict?

giorgiosironi: I'll annotate that for a follow-up

giorgiosironi: (needs testing)

PR #211, opened 2018-06-27
ELPP-3738 Add research-communication support

thewilkybarkid:


thewilkybarkid:
The test fails because it compares with data from `elife/api`, which hasn't yet been updated.

PR #224, opened 2018-11-12
Re-index

nlisgo:
This is required because of a necessary recovery of Journal CMS which resulted in the loss of one blog article.


giorgiosironi:
Follow-up: https://github.com/elifesciences/issues/issues/4601

PR #226, opened 2019-05-07
Rate limit all for= search queries

giorgiosironi:
Assuming Kong rate limiting works well (which is a leap), trying to use it to shut off the crawlers that are overloading this service.

### DIFF
--- a/src/Search/Kernel.php
+++ b/src/Search/Kernel.php
@@ -585,6 +585,10 @@ final class Kernel implements MinimalKernel
             $response->headers->set('X-Kong-Limit', 'highpages=1');
         }
 
+        if ($request->query->get('for', null)) {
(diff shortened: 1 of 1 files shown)


giorgiosironi:
https://github.com/elifesciences/builder/commit/3f8459e59cd2fead27aaf0b54eb602b7d559052e necessary.

giorgiosironi:
I am able to get
```
< HTTP/1.1 429 
* Server nginx is not blacklisted
< Server: nginx
< Date: Tue, 07 May 2019 16:23:39 GMT
< Content-Length: 0
< Connection: keep-alive
< Access-Control-Allow-Origin: *
< X-RateLimit-Limit-highpages-second: 1
< X-RateLimit-Remaining-highpages-second: 0
```
so rate limiting seems workable as a solution.

PR #229, opened 2019-05-28
Fail when Gearman can't be reached

giorgiosironi:
Don't hide exceptions in connecting to Gearman.

Discovered in https://github.com/elifesciences/search-formula/pull/20

I wouldn't merge this now as we need a stable pipeline to test it in all environments, there may be corner cases lurking.

Edit: it also probably doesn't help for the "connection hanging" scenario.


lsh-0:
oooh - how does one get a 'draft' pull request? I'll have to explore.

Looks good so far, :+1: fail fast and noisily

lsh-0:
(I'm going to merge this, draft or not)

lsh-0:
This fixes the hanging behaviour for me, cheers. I now have actual errors and broken states I can work with :)

PR #231, opened 2019-07-24
Fix: dev config, changes gearman server from 'localhost' to '127.0.0.1'

lsh-0:
* changes gearman server from 'localhost' to '127.0.0.1'

stops console from hanging in vagrant using Ubuntu 16.04. 

See: https://github.com/elifesciences/search-formula/pull/20

### DIFF
--- a/config/dev.php
+++ b/config/dev.php
@@ -7,7 +7,7 @@ return [
     'validate' => true,
     'api_url' => 'http://localhost:8080',
     'ttl' => 0,
-    'gearman_servers' => ['localhost'],
+    'gearman_servers' => ['127.0.0.1'],
     'gearman_auto_restart' => false,
     'elastic_force_sync' => true,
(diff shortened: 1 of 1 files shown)


lsh-0:
bit of a weird one here. `gearman` is listening on '127.0.0.1' for connections, and `./bin/console --list --env=dev` will successfully connect to gearman, as gearman logs:

> Accepted connection from 127.0.0.1:49734

but then the command hangs indefinitely, screwing up highstate. 

The gearman server itself doesn't seem to care if it's running as '127.0.0.1' or 'localhost' (unlike MySQL), it continues working normally, it's just the code in this project that is failing to resolve 'localhost' to '127.0.0.1' and finish execution.

lsh-0:
I'd like to merge this tomorrow and get the formula building on 16.04 if it's not too uncontroversial

PR #236, opened 2019-09-02
Fix: adds an exception to calling the smoke tests

lsh-0:
currently the smoke tests are *not* being called during the search-formula CI and the build is passing.

if they were being called, then the `fresh-follower-node` variant would be failing because there is no ElasticSearch instance to connect to.

the 18.04 search-formula PR *does* use the smoke tests during CI and is currently failing because of this.

### DIFF
--- a/smoke_tests.sh
+++ b/smoke_tests.sh
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 set -ex
 
+hostname | grep 'fresh-follower-node' && {
+    echo "This is a follower node without an elasticsearch leader. Smoke tests cannot be run."
+    exit 0
+}
+
 attempts=3
 delay=10
-
-# temporary, to get 18.04 formula branch passing. remove once upgraded. this delay is incorporated there
-sleep 20 
 
 retry ./smoke_tests_app.sh $attempts $delay
 retry ./smoke_tests_elasticsearch.sh $attempts $delay


lsh-0:
this may not be the best home for this exception to occur

giorgiosironi:
I thought https://github.com/elifesciences/search/blob/develop/Jenkinsfile.prod#L12 would always fail then; but not true as in that case there _is_ an ElasticSearch to connect to, it's just on node 1 rather than local.

smoke_tests.sh
```diff
@@ -1,11 +1,13 @@
...
+    exit 0
+}
+
 attempts=3
 delay=10
 
-# temporary, to get 18.04 formula branch passing. remove once upgraded. this delay is incorporated there
-sleep 20 
-
 retry ./smoke_tests_app.sh $attempts $delay
```
giorgiosironi: I think both of these would be failing as even the PHP side would try to connect to a non-existent ElasticSearch

smoke_tests.sh
```diff
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 set -ex
 
+hostname | grep 'fresh-follower-node' && {
```
giorgiosironi: just checked this is alright:
```
[2019-09-02T05:06:11.585Z] [54.234.147.38] out: [INFO    ] Executing state cmd.run for [echo pr-30-fresh-follower-node--search.elife.internal > /etc/hostname
```

smoke_tests.sh
lsh-0: yes indeed, it never gets to the elasticsearch set of smoketests as the app attempts to connect to elasticsearch even for the `/ping` endpoint

PR #237, opened 2019-09-03
changes gearman_server from hostname to '127.0.0.1' in all cases

lsh-0:
I think 'localhost' works as well on ec2 (as demonstrated on `--ci`) but not on vagrant, where it just hangs. I don't recall if we got the hanging behaviour in CI 

at time of writing end2end is [currently in a broken state](https://github.com/elifesciences/issues/issues/4789) because of this

### DIFF
--- a/config/ci.php
+++ b/config/ci.php
@@ -7,7 +7,7 @@ return [
     'api_url' => 'http://localhost:8080',
     'ttl' => 0,
     'elastic_force_sync' => true,
-    'gearman_servers' => ['localhost'],
+    'gearman_servers' => ['127.0.0.1'],
     'elastic_logging' => true,
     'gearman_auto_restart' => false,
     'aws' => [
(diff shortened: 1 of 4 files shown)


lsh-0:
oh - I think gearman is only running on the leader, in which case my changes are pure nonsense

lsh-0:
I have a better solution that doesn't involve modifying the search app. Closing

lsh-0:
> The assumption is correct, Gearman runs only on the leader so we cannot change the .elife.internal values here. You can change the localhost ones like ci.php if you want though, as there is a single node there

didn't see your comment there. My (incorrect) understanding was that gearman was a sort job queue for php because php doesn't do parallelism so well (a bit like python).

giorgiosironi:
> a sort of job queue for php

Yes, but the Gearman daemon is the server part so it is a single process like Elasticsearch here. There are multiple workers pulling jobs from it, but not completely sure if they run on both nodes or just on the leader.

PR #243, opened 2020-02-13
Use RDS companion article

thewilkybarkid:
https://github.com/elifesciences/issues/issues/5421


thewilkybarkid:
No tests for `ResearchArticleWorkflow`? 😨

config/dev.php
```diff
@@ -17,4 +17,12 @@
         'region' => 'us-east-1',
         'endpoint' => 'http://localhost:4100',
     ],
+    'feature_rds' => false,
+    'rds_articles' => [
+        '30274' => [
+            'date' => '2020-01-29T19:18:00Z',
+            'display' => 'https://hub.stenci.la/elife/30274/main/?elife',
+            'download' => 'https://hub.stenci.la/elife/30274/files/download/article-30274.md',
```
giorgiosironi: `download` should not be necessary?

thewilkybarkid: No, but trying to share the same config as Journal.

config/dev.php
```diff
@@ -17,4 +17,12 @@
         'region' => 'us-east-1',
         'endpoint' => 'http://localhost:4100',
     ],
+    'feature_rds' => false,
```
giorgiosironi: This `config/*.php` files will really need to be moved into pillars for ease of editing by production.

src/Search/Workflow/ResearchArticleWorkflow.php
```diff
@@ -132,7 +135,11 @@ public function index(ArticleVersion $article) : array
             'value' => json_encode($articleObject->dataSets ?? '[]'),
         ];
 
-        $sortDate = $article->getStatusDate();
```
giorgiosironi: seems we don't use other dates (like `versionDate`), so this should be the only entry point for articles

PR #249, opened 2020-02-18
Remove environment arguments

giorgiosironi:
Enables https://github.com/elifesciences/elife-alfred-formula/pull/33

The application doesn't need to know about the name of the environment anymore, as all options are duly passed in via `config.php`.

### DIFF
--- a/src/Search/Console.php
+++ b/src/Search/Console.php
@@ -258,6 +258,7 @@ final class Console
         // Some annotations
         Register::registerLoader();
 
+        // TODO: remove when it is *never* passed in by the formula or anything else
         $this->console->getDefinition()->addOption(new InputOption('--env', '-e', InputOption::VALUE_OPTIONAL, 'The Environment name. Deprecated and not used', 'dev'));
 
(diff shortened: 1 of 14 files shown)


src/Search/Console.php
```diff
@@ -258,6 +258,7 @@ public function __construct(Application $console, Kernel $kernel)
         // Some annotations
         Register::registerLoader();
 
+        // TODO: remove when it is *never* passed in by the formula or anything else
         $this->console->getDefinition()->addOption(new InputOption('--env', '-e', InputOption::VALUE_OPTIONAL, 'The Environment name. Deprecated and not used', 'dev'));
```
thewilkybarkid: > Deprecated and not used

Which is it? 😅

giorgiosironi: Deprecated to pass it in, because it is not used (but would explode if you pass it in and this definition is not here, IIRC)

thewilkybarkid: Deprecation preserves BC. Though maybe it was never used anyway.

PR #261, opened 2020-06-01
improved polling for the reindex process

lsh-0:
@nlisgo , this is a change to the polling script used in the Alfred 'process-search-reindex' task. It's why the task fails but the new index is successfully created.

The change introduces an error threshold that must be crossed before a worker can be said to definitely not exist. In this case there are a maximum of fifteen checks, one check every tenth of a second. This granularity with a single worker lets us see each each job being consumed. I've not seen it go past ten checks yet so fifteen should be safe.

### DIFF
--- a/bin/wait-for-empty-queues
+++ b/bin/wait-for-empty-queues
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 set -e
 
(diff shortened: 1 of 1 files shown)


lsh-0:
we'll run that Alfred process once it's deployed to `continuumtest` and `prod` (where it failed last time).

PR #263, opened 2020-06-04
wait-for-empty-queues, fixes

lsh-0:
~prod encounters a job that takes longer than almost everything else~

`prod` is doing something that causes the queue watcher to die or restart.

fyi @nlisgo

### DIFF
--- a/bin/wait-for-empty-queues
+++ b/bin/wait-for-empty-queues
@@ -6,8 +6,12 @@ set -e
 # more workers also means more CPU usage which means less CPU available for serving traffic.
 
 # how many attempts per-loop should be made to detect workers?
-no_workers_threshold=3000 # 3000 attempts with a .10s interval is five minutes.
-no_workers_attempt=0 # which attempt are we on now?
+num_workers_threshold=3000 # 3000 attempts with a .10s interval is five minutes.
+num_workers_attempt=0 # which attempt are we on now?
+
+# same again, but for the sqs watcher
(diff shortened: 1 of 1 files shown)


lsh-0:
I've instead opened a PR bumping the number of workers. We'll see how well that works with the current 5 second timeout.

lsh-0:
`prod` is looking promising and has made it further than usual.

I'm seeing these in the syslog though:

```
Jun 16 03:09:52 prod--search.elife.internal systemd[1]: search-gearman-worker@0.service: Service hold-off time over, scheduling restart.
Jun 16 03:09:52 prod--search.elife.internal systemd[1]: search-gearman-worker@0.service: Scheduled restart job, restart counter is at 3.
Jun 16 03:09:52 prod--search.elife.internal systemd[1]: Stopped "search-gearman-worker".
Jun 16 03:09:52 prod--search.elife.internal systemd[1]: Started "search-gearman-worker".
```

and this:

```
Jun 16 03:12:48 prod--search.elife.internal php[29991]: Notice: Undefined property: stdClass::$name in /srv/search/src/Search/Workflow/ResearchArticleWorkflow.php on line 104
```

and this:

```
Jun 16 03:14:31 prod--search.elife.internal php[32280]: Notice: Array to string conversion in /srv/search/src/Search/Workflow/Blocks.php on line 26
Jun 16 03:14:31 prod--search.elife.internal php[32280]: Notice: Array to string conversion in /srv/search/src/Search/Workflow/Blocks.php on line 26
```

PR #278, opened 2020-09-15
removes 'feature_rds' flag

lsh-0:


src/Search/Kernel.php
```diff
@@ -434,8 +433,7 @@ public function dependencies(Application $app)
                 $app['console.gearman.task_driver'],
                 $app['elastic.client.write'],
                 $app['validator'],
-                $app['logger'],
-                $app['config']['feature_rds'] ? $app['config']['rds_articles'] : []
```
nlisgo, line 438: The condition needs to be removed but the parameter is still expected. I will followup with a commit.

nlisgo (changes requested):

nlisgo (approved):

nlisgo:
@lsh-0 It has been a long time since CI has been green for any PR's

I will leave you to merge this in.

lsh-0:
> I will leave you to merge this in.

I'm more inclined to close it now to be honest.

It came up in Giorgio's PR fix, but now that RDS articles are properly integrated and are simply regular articles, do we need special handling for them in search at all? Can we just strip out all special handling of RDS articles? I can do this myself if it's the right thing to do.

lsh-0:
> @lsh-0 It has been a long time since CI has been green for any PR's

green now

PR #279, opened 2020-09-16
Swap RDS for ERA in code

nlisgo:


src/Search/Workflow/ResearchArticleWorkflow.php
```diff
@@ -80,7 +80,7 @@ public function validate(ArticleVersion $article) : ArticleVersion
                     'validation_error' => $this->validator->getLastError()->getMessage(),
                 ]
             );
-            throw new InvalidWorkflow('ResearchArticle<'.$article->getId().'> cannot be trasformed into a valid search result.');
+            throw new InvalidWorkflow('ResearchArticle<'.$article->getId().'> cannot be transformed into a valid search result.');
```
nlisgo, line 83: This is unrelated but spotted it while the file was open!

PR #284, opened 2020-10-06
Make reindex-rds test script start its own workers

giorgiosironi:
https://github.com/elifesciences/issues/issues/6173


giorgiosironi:
I would question the value of this test step as-is though, as the articles fail with a 404 request from `api-dummy` and are simply ignored. It's a smoke test that checks articles being reindexed don't make the workers crash, but doesn't guarantee the ElasticSearch index changes at all.

lsh-0:
is that `search.patch` file necessary? I'm going to remove it.

> I would question the value of this test step as-is though, as the articles fail with a 404 request from api-dummy and are simply ignored.

Indeed, the problem came to my attention because the `feature_rds` flag was being removed so the need for separate behaviour for 'rds' didn't seem necessary. I'll raise it with @nlisgo and see what he says.

PR #294, opened 2021-02-04
dropped log message about extant index from 'error' to 'warning'

lsh-0:
this will help remove the final bit of noise from the daily error logs we get from search--prod.


lsh-0:
merging this

PR #308, opened 2021-03-05
Check that all results in search validate against schema in CI

nlisgo:
This is a valuable CI check ahead of work that we wish to do on search.


src/Search/Console.php
```diff
@@ -326,11 +335,6 @@ public function __construct(Application $console, Kernel $kernel)
         }
     }
 
-    private function path($path = '')
```
nlisgo, line 329: This wasn't doing anything

NuclearRedeye (approved):
I've run over the code line by line, seems OK to me.

PR #311, opened 2021-03-16
Increase timeout thresholds

nlisgo:


nlisgo:
We are going to need a different approach here. Increasing the timeout on search will not reduce the errors that we are experiencing on journal unless we subsequently increase the timeouts over there. Currently we have a global timeout on journal but I may want to increase the timeout just for the search page on journal.

PR #317, opened 2021-04-08
Allow the content snippet to be written to elasticsearch

nlisgo:
Before switching to using a stored snippet rather than preparing a snippet we need to ensure that all the content currently in elasticsearch has a snippet available. We can achieve this by deploying this code and performing a reindex.

After determining that all content has a snippet we can then followup with https://github.com/elifesciences/search/pull/313

### DIFF
--- a/src/Search/Api/Elasticsearch/Command/resources/_default_.yaml
+++ b/src/Search/Api/Elasticsearch/Command/resources/_default_.yaml
@@ -44,3 +44,7 @@ _default_:
       type: date
       format: strict_date_optional_time
       index: not_analyzed
+
+    snippet:
+      type: string
+      index: not_analyzed
--- a/src/Search/Api/Elasticsearch/Command/resources/_default_.yaml
+++ b/src/Search/Api/Elasticsearch/Command/resources/_default_.yaml
@@ -46,5 +46,11 @@ _default_:
       index: not_analyzed
 
     snippet:
-      type: string
-      index: not_analyzed
+      type: object
+      properties:
+        format:
+          type: string
+          index: not_analyzed
+        value:
+          type: string
+          index: not_analyzed
--- a/src/Search/Workflow/CollectionWorkflow.php
+++ b/src/Search/Workflow/CollectionWorkflow.php
@@ -86,6 +86,11 @@ final class CollectionWorkflow implements Workflow
         $this->logger->debug('Collection<'.$collection->getId().'> Indexing '.$collection->getTitle());
         // Normalized fields.
         $collectionObject = json_decode($this->serialize($collection));
+        $collectionObject->snippet = $this->serializer->normalize(
+            $collection,
+            null,
+            ['snippet' => true, 'type' => true]
+        );
         $this->addSortDate($collectionObject, $collection->getPublishedDate());
 
         // Return.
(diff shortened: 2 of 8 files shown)


src/Search/Workflow/JsonSerializeTransport.php
```diff
@@ -44,4 +44,25 @@ public function serialize($article) : string
...
     }
+
+    public function snippet($article) : array
+    {
+        if (
+            !isset($this->serializer) ||
+            null === $this->serializer ||
+            !$this->serializer instanceof Serializer
+        ) {
+            throw new LogicException('You must inject API SDK serializer for this to work (property: $serializer missing.)');
```
giorgiosironi, line 55: I see this check is replicated from the other methods. Conservative to keep it

src/Search/Workflow/JsonSerializeTransport.php
```diff
@@ -44,4 +44,25 @@ public function serialize($article) : string
 
         return self::$cache[$key];
     }
+
+    public function snippet($article) : array
```
giorgiosironi, line 48: is this always an article or is it generic code? `$item`?

src/Search/Workflow/CollectionWorkflow.php
```diff
@@ -86,6 +86,7 @@ public function index(Collection $collection) : array
         $this->logger->debug('Collection<'.$collection->getId().'> Indexing '.$collection->getTitle());
         // Normalized fields.
         $collectionObject = json_decode($this->serialize($collection));
+        $collectionObject->snippet = ['format' => 'json', 'value' => json_encode($this->snippet($collection))];
```
giorgiosironi, line 89: These calls are all consistent with each other :+1:

src/Search/Api/Elasticsearch/Command/resources/_default_.yaml
```diff
@@ -44,3 +44,13 @@ _default_:
       type: date
       format: strict_date_optional_time
       index: not_analyzed
+
+    snippet:
+      type: object
+      properties:
+        format:
+          type: string
+          index: not_analyzed
```
giorgiosironi, line 53: seems like the right intent to prevent as much work from Elasticsearch as possible. The index will already be heavy

nlisgo:
Thanks for the feedback @giorgiosironi

src/Search/Workflow/JsonSerializeTransport.php
```diff
@@ -11,15 +11,28 @@ trait JsonSerializeTransport
...
+                !isset($this->serializer) ||
+                null === $this->serializer ||
+                !$this->serializer instanceof Serializer
+            ) {
+                throw new LogicException('You must inject API SDK serializer for this to work (property: $serializer missing.)');
+            }
+            $checked = true;
         }
+
+        return $checked;
```
nlisgo, line 29: Thought this was a better way of performing the check just once.

src/Search/Workflow/JsonSerializeTransport.php
```diff
@@ -30,18 +43,24 @@ public function deserialize(string $json)
...
         }
 
         return self::$cache[$key];
     }
+
+    public function snippet($article) : array
+    {
+        $this->checkSerializer();
+
+        return $this->serializer->normalize(
```
nlisgo, line 60: Removed the cache code here because we don't need to read this value back.

src/Search/Workflow/JsonSerializeTransport.php
lsh-0: Seems unnecessary to me. The checks being done are cheap, probably taking less than a single ms. I would recommend reverting it.

nlisgo: It wasn't really about saving time but removing redundancy. Will reinstate, keeping it in it's own function but running on each call.

PR #324, opened 2021-05-13
Apply search word limit and log searches that exceed it

nlisgo:


NuclearRedeye (approved):
I'm no PHP expert, so can't comment on the specifics of the code but functionally it covers everything that we discussed and agreed. Happy to try and give these changes a test locally tomorrow before they are merged if that would help?

nlisgo:
Have added a test for the new `applyWordLimit` method on `ElasticQueryBuilder`.

Marking "Ready for Review"

NuclearRedeye:
And spent some time testing this locally today and it's working well for me. Even after truncating the results I'm still getting the exact result expected so from my perspective I don't think this will cause end users too much issue except for the occasional edge case.

PR #325, opened 2021-05-19
Introduce a performance test to ci

nlisgo:
Ultimately end2end might be the best place for this. The performance test in place is currently being performed in around 30 seconds and is only occasionally triggered any failed requests.

The script extracts the total time and failed requests values from the output of ApacheBench and this should allow us to set thresholds for these values and potentially fail builds if they exceed the threshold.

### DIFF
--- a/bin/ci-import
+++ b/bin/ci-import
@@ -18,3 +18,4 @@ echo "Finished launching import"
 
 . bin/ci-wait-steady-state
 . bin/ci-expected-results
+. bin/ci-performance
--- a/bin/ci-import
+++ b/bin/ci-import
@@ -18,4 +18,3 @@ echo "Finished launching import"
 
 . bin/ci-wait-steady-state
 . bin/ci-expected-results
-. bin/ci-performance
--- a/bin/ci-performance
+++ b/bin/ci-performance
@@ -1,4 +1,12 @@
 #!/bin/bash
 set -e
 
-ab -n 100 -c 5 "http://search--vagrant/search?for=Updates%20to%20the%20zoonotic%20niche%20map%20of%20Ebola%20virus%20disease%20in%20Africa&page=1&per-page=10&sort=relevance&order=desc&use-date=default"
+bench=$(ab -n 100 -c 5 "http://search--vagrant/search?for=Updates%20to%20the%20zoonotic%20niche%20map%20of%20Ebola%20virus%20disease%20in%20Africa&page=1&per-page=10&sort=relevance&order=desc&use-date=default")
+
+echo "$bench\n"
+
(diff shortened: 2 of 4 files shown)


nlisgo:
There was an issue when using localhost with connection refused

https://stackoverflow.com/questions/7938869/ab-is-erroring-out-with-apr-socket-recv-connection-refused-61/8114719

NuclearRedeye (approved):
I tested locally and it works well. One improvement that I considered was if we could randomise the search term each time (so we avoid any caching that elasticsearch might do) and hence have a more realistic test. I found [this blog post](https://chrismiles.livejournal.com/21720.html) but sadly the link to the patch is broken.

As a starting point though, this is a good foundation to run on.

lsh-0:
It's important to be clear in what is being tested and what your expectations are. There are no expectations in `ci-performance`, nothing is being tested.

I would recommend getting a baseline for throughput/time period on CI and enforcing that. 

For example: "100 requests/5 seconds", or whatever /search in `ci` and `continuumtest` and `prod` can handle. NewRelic can give us some insight into the load `search--prod` current sees.

This can be achieved by setting the `-t` (timelimit) parameter in `ab`. Just like testing code coverage we set it low to begin with and then ratchet it upwards as our results improve. This will ensure regressions in basic performance don't occur.

>  One improvement that I considered was if we could randomise the search term each time (so we avoid any caching that elasticsearch might do)

These are 'performance' tests, we're interested in throughput. If you start randomising search terms then it becomes [a different type of test](https://en.wikipedia.org/wiki/Fuzzing). Bypassing any caching mechanism is a legitimate concern though. If we find ourselves having to debug performance tests then we want our future selves to have an easy time of it; we'll want those search terms to have been, at the very least, deterministic and preferably taken from a small pool of possible search terms. Cache busting might have to happen in some other search parameter, like page number or order or num results etc. 

I think this is a step in the right direction. General purpose load testing and fuzzing on a non-local machine are good ideas as well and something I've wanted for a long time. Giorgio and I had [siege](https://www.joedog.org/siege-home/) installed on Alfred once upon a time with the intent to incorporate load testing, I think that idea petered out in the face of the quick development that was happening but it's still a good idea.

nlisgo:
@NuclearRedeye and @lsh-0 appreciate the feedback.

I will introduce a threshold. As suggested, we will set a low bar initially. Ideally I would like this to be used for monitoring and see how it appears over time. But I don't think we can achieve this other than on end2end.

I accept it might be interesting to throw some different requests (not necessarily random) but with the small amount of data we have in api-dummy it is important that we hit requests that return results as I suspect that it is the post filter boostings that are expensive.

I will introduce a threshold and then merge this in.

lsh-0:
sounds good

PR #362, opened 2022-01-25
Escape reserved characters in search queries

nlisgo:


src/Search/Api/Elasticsearch/ElasticQueryBuilder.php
```diff
@@ -180,6 +180,12 @@ public function searchFor(string $string) : QueryBuilder
         return $this;
     }
 
+    private function escapeReservedChars(string $string) : string
+    {
+        // Reintroduce support for some of the characters when they can be appropriately handled.
+        return preg_replace(['@[><]@', '@[-+=&|!(){}[\]^"~*?:\\\/]@'], ['', addslashes('\\$0')], $string);
```
lsh-0, line 186: if some of this was copied from stackoverflow ([for example](https://stackoverflow.com/questions/33845230/escape-elasticsearch-special-characters-in-php/68389722#68389722)) you should attribute it/link to where you pulled it from.

I don't think we have to be worried about licencing issues when sourcing from SO but solutions now often become superseded later and ... it's just polite.

I found that regex by following links in this ticket: 
https://github.com/elastic/elasticsearch-php/issues/620

which also has this helpful link (couldn't find an Opensearch version):
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters

could I get that link pasted above that regex? It would help explain what all those things are you're removing or escaping.

And I'm not entirely sure what the result of `addslashes('\\$0')` looks like but hardcoding the result might be clearer.

nlisgo: > I don't think we have to be worried about licencing issues when sourcing from SO but solutions now often become superseded later and ... it's just polite.

I agree. Will add this.

bin/ci-expected-results
```diff
@@ -70,6 +70,16 @@ if [ "$actual_author_non_authorline_title" != "$expected_author_non_authorline_t
     exit 2
 fi
 
+actual_total=$(curl -v "localhost/search?for=rheumatoid%20arthritis%20(RA)&page=1&per-page=10&sort=relevance&order=desc&use-date=default" | jq ".total")
```
NuclearRedeye, line 73: As `actual_total` is used earlier in the script, I'd consider renaming this to something more representative.

bin/ci-expected-results
```diff
@@ -70,6 +70,16 @@ if [ "$actual_author_non_authorline_title" != "$expected_author_non_authorline_t
...
+if [ ! "$actual_total" ]; then
```
NuclearRedeye, line 74: Is this checking to see if it the value has been set or not? If so, a better approach would be...

`if [ -z "$actual_total" ]; then`

Which both checks that the variable exists and has been set to anything other than empty.

bin/ci-expected-results
```diff
@@ -70,6 +70,16 @@ if [ "$actual_author_non_authorline_title" != "$expected_author_non_authorline_t
...
+    actual_total=-1
+fi
+echo "CHECK: Search with reserved character"
+if [ ! "$actual_total" -gte 0 ]; then
```
NuclearRedeye, line 78: You could use `-le` rather than `! -gte`

NuclearRedeye: Or `-eq` if you need 0 on the nose

NuclearRedeye (approved):
Some optional items you could address

bin/ci-expected-results
nlisgo: Would it be `if [ ! -z "$actual_total" ]; then`?

NuclearRedeye: Sorry yes, for the inverse where it isn't set or empty you need the `!`

lsh-0: shellcheck prefers `-n` to `! -z`. It's a great little tool: https://www.shellcheck.net

NuclearRedeye: I've not come across ShellCheck before, what a great tool. Bookmark and installed.

PR #366, opened 2022-02-01
fix: project tests are failing in ci

NuclearRedeye:


lsh-0 (commented):
just to summarise for my own benefit:

* new script `ci-empty-queue-and-wait`
* replaces index `elife_search` with `ci-import`
* replaces index `elife_search_temporary_lifecycle_test` with `ci-lifecycle`
* replaces index `elife_search_$(date +%Y%m%d%H%M%S)` with `ci-reindex`
* more polling for specific states in general with hard fails if they're not reached (`exit 1`)
* more consistent naming for `ci` scripts

I keep forgetting `elife-search--ci` doesn't use systemctl to handle the workers+watchers (sorry).

I think overall the changes are fine. The hard failures when states aren't reached will make this brittle, but strict and brittle is better than lax, ambiguous and confusing. And besides, we *should* be able to consistently empty those bloody queues without hassle.

The removal of the `elife_search` index *might* require a change to the formula, I'm not sure, I'll need to double check. The older elasticsearch work ensured that index existed and I don't recall if I carried it over for opensearch.

I'll add some code comments but there's no deal breakers here for me - assuming it fixes the problem ;) Perhaps turning of Jenkins' `process-stop-stacks` and running this PR several times one after the other might give us more confidence. We often see the tests pass after the instance is restarted.

bin/ci-empty-queue-and-wait
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
```
lsh-0, line 2: no `set -e` ?

It's preferred but not mandatory. If there are commands that may fail execution we should know which ones and how they might break. I see some `|| true` happening as well which typically mark such commands.

Given how the scripts are included and executed I suspect there might be some kind of flag inheritance happening but these scripts can also be run by themselves.

bin/ci-start-processes
```diff
@@ -1,7 +1,46 @@
+#!/bin/bash
+
```
lsh-0, line 2: another missing `set -e`, see previous comment. If it's deliberate that's fine

bin/reindex-rds
```diff
@@ -1,11 +1,8 @@
 #!/usr/bin/env bash
 set -e
-. bin/ci-process-bootstrap
-
-. bin/ci-start-processes
 
 ./bin/console rds:reindex
 
 echo "Waiting for empty queues"
```
lsh-0, line 6: perhaps stick this comment into `./bin/wait-for-empty-queues`

bin/ci-wait-steady-state
```diff
@@ -1,11 +1,23 @@
...
-fi
+#!/bin/bash
+
+echo -n "Waiting for queues to empty..."
+for i in $(seq 60); do
+  echo -n "."
+
+  sqs_queue=$(bin/console queue:count || true)
+  gearman_queue=$(gearadmin --show-jobs | wc -l || true)
+  if [ "$sqs_queue" -eq "0" ] && [ "$gearman_queue" -eq "1" ]; then
```
lsh-0, line 9: magic literal of `1`

why are we waiting for a value of 1 and not 0 (like `$sqs_queue`) when we're emptying queues? If it's deliberate please add a comment

bin/ci-wait-steady-state
```diff
@@ -1,11 +1,23 @@
...
+
+  if [ "$i" -eq "60" ]; then
+    echo "FAILED"
+    exit 1
+  fi
+
+  sleep 1
+done
+
+# Give the index some time to make the content available
```
lsh-0, line 22: by 'the index' do you mean opensearch?

bin/ci-reindex-rds
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -e
+. bin/ci-process-bootstrap
+
+. bin/ci-start-processes
+
+./bin/console rds:reindex
+
+echo "Waiting for empty queues"
```
lsh-0, line 9: redundant: `./bin/ci-wait-steady-state` does an `echo -n "Waiting for queues to empty..."` immediately.

bin/reindex-rds
NuclearRedeye: Done. The bigger scare for me is that this script was prior to this PR being run on both CI and Production. In this instance, running this on production would spawn a watcher and a worker bypassing `systemd` and if the script fails there would be dangling processes left behind.
This change should resolve that issue, although it assumes that a worker and watcher will already running.

bin/ci-wait-steady-state
NuclearRedeye: Good spot, it is intentional so have added a comment to capture why.

bin/ci-reindex
```diff
@@ -2,21 +2,19 @@
...
 
-./bin/console queue:clean
-# TODO: poll for queue:count == 0, with timeout
-sleep 5
+# Clean the queues
+./bin/ci-empty-queue-and-wait
 
-index_name=elife_search_$(date +%Y%m%d%H%M%S)
-./bin/console search:setup --index="$index_name" --delete
+./bin/console search:setup --index="ci-reindex" --delete
```
nlisgo, line 8: I've no objection to removing the date and time from the index name. This could be useful if rerunning locally and debugging. Not necessary though.

project_tests.sh
```diff
@@ -15,4 +15,4 @@ echo "Reindexing api-dummy"
 bin/ci-reindex
 
 echo "Reindexing RDS articles"
-bin/reindex-rds
+bin/ci-reindex-rds
```
nlisgo, line 18: Can we make changes to `bin/reindex-rds` to not have to support 2 files?

nlisgo (approved):
Some comments but nothing blocking.

project_tests.sh
NuclearRedeye: With some further work possibly, but the reason they need to be separate now is because the mechanisms for starting the workers and queue watchers is different between the CI and Production environments, and in this case I discovered that the CI script is being used by prod which could result in an edge case of content being inserted into the wrong index, similar to what I'm able to see locally.
As such, I split the scripts out for now and cleaned the production one up to not start the processes manually and instead rely on the fact that `systemd` has already started them.

bin/ci-reindex
NuclearRedeye: These changes should stop indexes accumulating as the scripts would never delete them. Insures that the tests start from a more consistent baseline.

lsh-0: *ensures

There is a ./bin/console command called `index:delete:unused` that will "Delete all unused indexes". I tried running it recently and didn't see any change.

Handling accumulating indices is easy but debugging is typically hard. See how you go @nlisgo and we can always add a datestamp back.

project_tests.sh
lsh-0: +1 , these scripts are already spaghetti and trading a little duplicated code for more clarity isn't so bad.

PR #369, opened 2022-02-17
fix: set result window size when creating an index

NuclearRedeye:


NuclearRedeye:
@nlisgo I've created this draft PR which shows how to configure the result window size which resolves elifesciences/issues#7211, but this needs some further work to make the setting configurable rather than hard coded which I'm not all that familiar with and think you might be best placed to advise on.

PR #373, opened 2022-02-23
fix: add limit to reject requests that exceed the maximum result window size

NuclearRedeye:


nlisgo:
Reviewing this. Will need to give it more time in order to weigh on this. Will look at this today.

nlisgo (approved):

nlisgo:
@NuclearRedeye over to you.

PR #375, opened 2022-03-14
composer.json, relaxes php version constraint.

lsh-0:
so it can be installed under php 7.4

fyi @NuclearRedeye @nlisgo

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,10 @@
         }
     },
     "require": {
-        "php": "~7.2.0",
+        "php": ">=7.2.9,<=7.4.9",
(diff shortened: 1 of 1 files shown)


lsh-0:
I *suspect* I may not be able to get away with this, as it might allow packages not supported on php 7.2. It worked for php 7.4 on `search` in vagrant

lsh-0:
yeah, did some more exploration and you can't. I have a better way, I'll open another PR

lsh-0:
yeah, did some more exploration and you can't. I have a better way, I'll open another PR

PR #384, opened 2022-09-08
add support for version 2

smoqadam:


composer.json
```diff
@@ -21,7 +21,7 @@
         "elife/api": "^2.0",
         "elife/api-client": "dev-master",
         "elife/api-problem": "^1.1",
-        "elife/api-sdk": "dev-master",
```
nlisgo, line 24: We can still rely upon `dev-master` in order to get this through

tests/src/Search/Web/TypeTest.php
```diff
@@ -25,7 +25,7 @@ public function typeProvider() : Traversable
             'application/vnd.elife.search+json' => 200,
             'application/vnd.elife.search+json; version=0' => 406,
             'application/vnd.elife.search+json; version=1' => 200,
-            'application/vnd.elife.search+json; version=2' => 406,
+            'application/vnd.elife.search+json; version=2' => 200,
```
nlisgo, line 28: We should add a test for version 3 to return 406.

nlisgo (changes requested):
We also need a test to verify that the response header contains the appropriate Content-Type  value.

tests/src/Search/Web/TypeTest.php
smoqadam: @nlisgo added.

.docker/goaws.yaml
```diff
@@ -0,0 +1,10 @@
+Dev:                                # Another environment
+  Host: localhost
+  Port: 4100
+  # SqsPort: 9324
+  # SnsPort: 9292
+  Region: us-east-1
+  LogToFile: true
+  LogFile: ./goaws_messages.log
+  Queues:
+    - Name: search--dev
```
nlisgo, line 10: Doesn't belong in this PR.

docker-compose.yml
```diff
@@ -0,0 +1,70 @@
+version: '3'
```
nlisgo, line 1: Doesn't belong in this PR.

src/Search/Kernel.php
```diff
@@ -399,10 +399,13 @@ public function dependencies(Application $app)
                 $config['endpoint'] = $app['config']['aws']['endpoint'];
             }
             if (!isset($app['config']['aws']['credential_file']) || false === $app['config']['aws']['credential_file']) {
-                $config['credentials'] = [
-                    'key' => $app['config']['aws']['key'],
-                    'secret' => $app['config']['aws']['secret'],
-                ];
+                $config['credentials'] = false;
```
nlisgo, line 402: Doesn't belong in this PR.

tests/src/Search/Web/TypeTest.php
```diff
@@ -10,27 +11,56 @@ final class TypeTest extends ElasticTestCase
...
         foreach ($types as $type => $statusCode) {
             yield $type => [$type, $statusCode];
         }
     }
+
+    /**
+     * @test
+     * @dataProvider contentTypeProvider
+     */
+    public function testContentType($type, $contentType)
```
nlisgo, line 40: add `string` type hints

tests/src/Search/Web/TypeTest.php
```diff
@@ -10,27 +11,56 @@ final class TypeTest extends ElasticTestCase
...
+    {
+        $response = $this->performRequest($type);
+        $this->assertSame($contentType, $response->headers->get('Content-Type'));
+    }
+
+    public function contentTypeProvider()
```
nlisgo, line 46: Should be `public function contentTypeProvider() : Traversable`

nlisgo (changes requested):

nlisgo:
I believe the error is being caused by the search response not being valid as reviewed-preprint is not present in the types array.

nlisgo:
Tests are green but the new `reviewed-preprint` types value will be introduce regardless of the version requested. This isn't ideal but I'm happy to merge in as long as we confirm that journal search page doesn't react in a strange way. Let's talk about this tomorrow.

nlisgo (approved):

PR #396, opened 2022-10-24
Update to phpunit 9.5

nlisgo:


scottaubrey:
We're now on phpunit 9.6

PR #401, opened 2022-11-23
Remove dependencies on api-sdk in workflow tests

smoqadam:


tests/src/Search/Workflow/WorkflowTestCase.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        foreach ($this->findSamples($this->getModel() ?? $model, $this->getVersion() ?? $version) as $name => $sample) {
+            $object = $this->getSerializer()->denormalize($sample, $this->getModelClass() ?? $modelClass);
+            yield $name => [$object];
+        }
+    }
+
+    final protected function findSamples(string $model, int $version) : Traversable
+    {
+        $samples = Finder::create()->files()->in(
```
nlisgo, line 77: We could perhaps just move findSamples logic into workflowProvider.

tests/src/Search/Workflow/WorkflowTestCase.php
```diff
@@ -0,0 +1,84 @@
...
+
+abstract class WorkflowTestCase extends PHPUnit_Framework_TestCase
+{
+    use AsyncAssert;
+    use HttpMocks;
+    use GetSerializer;
+    use GetValidator;
+
+    /**
+     * @var BlogArticleWorkflow
```
nlisgo, line 29: This should be `@var Workflow`. We could then also remove the use statement for BlogArticleWorkflow

nlisgo (approved):

PR #402, opened 2022-11-24
Dockerize the app

achirazi:


smoqadam (commented):
Since we renamed the `opensearch-node` to `opensearch` in docker-compose.yaml, we must rename it everywhere else such as other services in docker-compose.yaml and config.php.dist, etc.

achirazi:
> Since we renamed the `opensearch-node` to `opensearch` in docker-compose.yaml, we must rename it everywhere else such as other services in docker-compose.yaml and config.php.dist, etc.

Fixed it. I went through all the files. If I missed something, please add another comment to that specific line.

Are you happy with the README.md that I've added, or do I need to add more info?

smoqadam (approved):

composer.json
```diff
@@ -46,12 +46,13 @@
...
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/browser-kit": "^3.1",
-        "symfony/var-dumper": "^3.1"
+        "symfony/var-dumper": "^3.1",
+        "ext-gearman": "*"
     },
     "config": {
         "sort-packages": true,
         "allow-plugins": {
-            "mindplay/composer-locator": true
```
nlisgo, line 54: Reinstate this.

composer.json
```diff
@@ -46,12 +46,13 @@
         "phpunit/phpunit": "^5.5",
         "squizlabs/php_codesniffer": "^3.5",
         "symfony/browser-kit": "^3.1",
-        "symfony/var-dumper": "^3.1"
+        "symfony/var-dumper": "^3.1",
+        "ext-gearman": "*"
```
nlisgo, line 50: Would expect this to introduce a change to composer.lock

README.md
```diff
@@ -0,0 +1,17 @@
+eLife Search
+=============
+
+This project is using elasticsearch to index content of the eLife articles and provide a full-text search for Journal.
```
nlisgo, line 4: It's important that we state that docker is just used to improve the developer experience and that to reliably recreate any issue experienced in ci or prod we should continue to use builder.

README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Installation
+------------
+
+1. Clone the project `git clone https://github.com/elifesciences/search.git`
+2. Run `docker-compose up --build`
+
+Important : Rename `config.php.dist` on local to `config.php` because config.php is in .gitignore.
```
nlisgo, line 12: This step should probably go before the docker command.

src/Search/Kernel.php
```diff
@@ -399,10 +400,10 @@ public function dependencies(Application $app)
                 $config['endpoint'] = $app['config']['aws']['endpoint'];
             }
             if (!isset($app['config']['aws']['credential_file']) || false === $app['config']['aws']['credential_file']) {
-                $config['credentials'] = [
-                    'key' => $app['config']['aws']['key'],
-                    'secret' => $app['config']['aws']['secret'],
-                ];
+                $config['credentials'] = new Credentials(
```
nlisgo, line 403: If not sure why this is necessary. Will investigate when I can access the app.

README.md
nlisgo: How should we access the search api?

http://localhost:????

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apt update && apt install libgearman-dev -y && pecl install gearman
+
+RUN docker-php-ext-enable gearman \
+    && docker-php-ext-configure pcntl --enable-pcntl \
+    &&  docker-php-ext-install pcntl
+
+COPY .docker/php.ini /usr/local/etc/php/php.ini
+
+WORKDIR /app
```
nlisgo, line 11: Since the Dockerfile and docker-compose.yaml are just needed for developer experience I think we should consider placing them in a subfolder or renaming them so they are not automatically picked up. We can still advise in the README how the docker container can be started. Let's discuss tomorrow.

README.md
achirazi: Added that in my last commit

composer.json
achirazi: Fixed in last commit

README.md
achirazi: Added step before docker command.

The search API from what @smoqadam shared with me is not accessible from browser.
We had to remove gearman-ui and opensearch-dashboard due to the fact that if we disable SSL for opensearch, the dashboard also needs to be configured to connect to opensearch without ssl (I don't know if that's possible).

Dockerfile
achirazi: Is there any other way to set up docker for development? 

I don't have enough info on search. I'm happy to insert any additional info inside Readme, just feel free to let me know what needs to be added.

src/Search/Kernel.php
achirazi: Reverted changes in last commit. @smoqadam told me that he will include this change in his current branch

Dockerfile
achirazi: Created the subfolder and added the file in there in my last commit

README.md
```diff
@@ -0,0 +1,20 @@
...
+This project is using elasticsearch to index content of the eLife articles and provide a full-text search for Journal.
+
+Important: Keep in mind that docker is just used to improve the developer experience. To reliably recreate any issue experienced in CI or Prod you should continue to use [builder](https://github.com/elifesciences/builder).
+
+Installation
+------------
+
+1. Clone the project `git clone https://github.com/elifesciences/search.git`
+2. Rename `config.php.dist` on local to `config.php` because config.php is in .gitignore.
+3. Run `docker-compose -f ./dev/docker-compose.yml up --build`
```
smoqadam, line 13: it's `docker-compose.yaml`

dev/docker-compose.yaml
```diff
@@ -0,0 +1,75 @@
...
+            - '4100:4100'
+        networks:
+            - opensearch-net
+    app:
+        build:
+            context: .
+            dockerfile: Dockerfile
+        volumes:
+            - ..:/app
+            - ./config.php:/app/config.php
```
nlisgo, line 58: I haven't confirmed that this works yet. I am on a train at the moment so I will try this out when I'm not on the move.

dev/.docker/goaws.yaml
```diff
@@ -0,0 +1,10 @@
+Dev:                                # Another environment
+  Host: goaws
+  Port: 4100
+  # SqsPort: 9324
+  # SnsPort: 9292
```
nlisgo, line 5: Why include these if we are not using them?

dev/.docker/goaws.yaml
```diff
@@ -0,0 +1,10 @@
+Dev:                                # Another environment
```
nlisgo, line 1: Get rid of `# Another environment` comment

achirazi: Fixed in last commit

dev/.docker/goaws.yaml
achirazi: Fixed in last commit

nlisgo:
I'm hitting this error:

```
Error response from daemon: Conflict. The container name "/goaws" is already in use by container "7ad115f130e3e42ab7609f50260337225eab0df9e9929e1691fc6a43315244d2". You have to remove (or rename) that container to be able to reuse that name.
```

On running: `docker compose -f dev/docker-compose.yaml up --build`

Is this the issue you are experiencing @smoqadam ?

README.md
```diff
@@ -0,0 +1,48 @@
...
+1. Go to `app` container:
+
+```bash 
+$ docker-compose -f dev/docker-compose.yaml exec app /bin/bash
+```
+
+2. Run following commands in order:
+
+```bash 
+$ bin/console queue:create
```
nlisgo, line 28: Getting the following error at this step: 

```
root@e0bbc1a8fb5e:/app# bin/console queue:create

Fatal error: Uncaught Elasticsearch\Common\Exceptions\NoNodesAvailableException: No alive nodes found in your cluster in /app/vendor/elasticsearch/elasticsearch/src/Elasticsearch/ConnectionPool/StaticNoPingConnectionPool.php:64
Stack trace:
#0 /app/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Transport.php(82): Elasticsearch\ConnectionPool\StaticNoPingConnectionPool->nextConnection()
#1 /app/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Transport.php(99): Elasticsearch\Transport->getConnection()
#2 /app/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php(297): Elasticsearch\Transport->performRequest('GET', '/key_value_stor...', Array, NULL, Array)
#3 /app/vendor/react/promise/src/FulfilledPromise.php(28): Elasticsearch\Connections\Connection->Elasticsearch\Connections\{closure}(Array)
#4 /app/vendor/ezimuel/ringphp/src/Future/CompletedFutureValue.php(55): React\Promise\FulfilledPromise->then(Object(Closure), NULL, NULL)
#5 /app/vendor/ezimuel/ringphp/src/Core.php(341): GuzzleHttp in /app/vendor/elasticsearch/elasticsearch/src/Elasticsearch/ConnectionPool/StaticNoPingConnectionPool.php on line 64
```

dev/docker-compose.yaml
nlisgo: This seems to be working as expected.

nlisgo:
> I'm hitting this error:
> 
> ```
> Error response from daemon: Conflict. The container name "/goaws" is already in use by container "7ad115f130e3e42ab7609f50260337225eab0df9e9929e1691fc6a43315244d2". You have to remove (or rename) that container to be able to reuse that name.
> ```
> 
> On running: `docker compose -f dev/docker-compose.yaml up --build`
> 
> Is this the issue you are experiencing @smoqadam ?

I removed the existing containers with the same name and no longer get this error.

README.md
smoqadam: @nlisgo It happens because it takes some time for elasticsearch node to be ready, if give it a minute or two, you won't see this error.

dev/docker-compose.yaml
smoqadam: What is the reason for putting `config.php.dist` into `dev` folder? If someone wants to run the app with `builder`, they need to copy it into the root folder. Anyhow, if we decide to keep it in `dev`, then, we should change this line to `- ./config.php.dist:/app/config.php`

README.md
```diff
@@ -9,8 +9,9 @@ To reliably recreate any issue experienced in CI or Prod you should continue to
 Important: Keep in mind that docker is just used to improve the developer experience.
 
 1. Clone the project `git clone https://github.com/elifesciences/search.git`
-2. Rename `/dev/config.php.dist` on local to `/dev/config.php` 
-3. Run `docker-compose -f dev/docker-compose.yaml up --build`
+2. Rename `/dev/config.php.dist` on local to `/dev/config.php`
+3. Run `mkdir logs && sudo chmod 777 ./logs`
```
nlisgo, line 13: We could add logs to dev folder with .gitkeep if needed? This would avoid the need to have this step but will try this out.

README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Important: Keep in mind that docker is just used to improve the developer experience.
+
+1. Clone the project `git clone https://github.com/elifesciences/search.git`
+2. Rename `/dev/config.php.dist` on local to `/dev/config.php`
+3. Run `mkdir logs && sudo chmod 777 ./logs`
+4. Run `docker-compose -f dev/docker-compose.yaml up --build`
+
+### Setup
+
```
nlisgo, line 17: Might be worth adding a step to make sure that we are ready for setup? For instance, making sure that opensearch is running/working?

nlisgo, line 17: Might be worth adding a step to make sure that we are ready for setup? For instance, making sure that opensearch is running/working?

nlisgo (approved):
Will create a followup ticket to address the issues that are causing `project_tests.sh` not to work.

We should also consider adding kibana as this is a powerful elasticsearch client.

PR #404, opened 2022-12-02
Check if current ArticleVersion is VOR and is a reviewed preprint

smoqadam:


nlisgo (changes requested):
We also need to refactor the following code which currently uses the yaml file: https://github.com/elifesciences/search/blob/develop/src/Search/Console.php#L350-L365

bin/ci-reindex-reviewed-preprints
```diff
@@ -4,6 +4,6 @@ set -e
 
 . bin/ci-start-processes
 
-./bin/console reviewedPreprint:reindex
+./bin/console queue:import ReviewedPreprints
```
nlisgo, line 7: As we are no longer using the reviewed preprints config we can get rid of the `reviewedPreprint:reindex` command as it duplicates the above command.

src/Search/Console.php
```diff
@@ -343,26 +341,15 @@ public function rdsReindexCommand(InputInterface $input, OutputInterface $output
...
-        $this->logger->info('Reviewed preprints added to indexing queue.');
-    }
-
     public function reviewedPreprintPurgeCommand(InputInterface $input, OutputInterface $output)
     {
         $this->logger->info('Purge reviewed preprints...');
         $ids = [];
-        foreach (array_keys($this->config['reviewed_preprints']) as $id) {
+        /** @var ApiSdk $sdk */
+        $sdk = $this->kernel->get('api.sdk');
```
nlisgo, line 349: Let's remove the need to call the api here and just purge all reviewed preprints from the elasticsearch. There may be a single command to do this or we can read them from the index and then delete each one in turn.

nlisgo (changes requested):
Nearly there just one other optimisation on the purge command. Thanks!

src/Search/Console.php
```diff
@@ -343,26 +342,15 @@ public function rdsReindexCommand(InputInterface $input, OutputInterface $output
...
-    }
-
     public function reviewedPreprintPurgeCommand(InputInterface $input, OutputInterface $output)
     {
         $this->logger->info('Purge reviewed preprints...');
         $ids = [];
-        foreach (array_keys($this->config['reviewed_preprints']) as $id) {
+        $query = new ElasticQueryBuilder($this->kernel->indexMetadata()->read());
+        $query = $query->whereType(['reviewed-preprint']);
+        $reviewedPreprints = $this->kernel->get('elastic.client.read')->searchDocuments($query->getRawQuery());
```
nlisgo, line 351: We should introduce a try catch here.

See: https://github.com/elifesciences/search/blob/develop/src/Search/Api/SearchController.php#L154-L168

src/Search/Console.php
```diff
@@ -343,26 +342,15 @@ public function rdsReindexCommand(InputInterface $input, OutputInterface $output
...
-            $this->enqueue('article', $id);
-            $this->logger->info("Queuing reviewed preprint $id");
-            $this->enqueue('reviewed-preprint', $id);
-            $ids[] = $id;
-        }
-        $output->writeln('Queued: '.implode(', ', $ids).' (reviewed-preprint and article)');
-        $this->logger->info('Reviewed preprints added to indexing queue.');
-    }
-
     public function reviewedPreprintPurgeCommand(InputInterface $input, OutputInterface $output)
```
nlisgo, line 345: We could create a follow up ticket to create a single command to allow purging by type.

smoqadam: Ticket has been created: https://github.com/elifesciences/issues/issues/7985

nlisgo:
There is a failure on CI with phpcs.

tests/src/Search/Workflow/ResearchArticleWorkflowTest.php
```diff
@@ -85,10 +86,10 @@ public function testRdsDateIsUsedAsTheSortDateWhenThereIsAnRdsArticle()
     public function testReviewedDateAndCurationLabelsWhenThereIsAReviewedPreprint()
```
nlisgo, line 86: We can remove this test as we are not decorating the snippet.

nlisgo (approved):
Pending removal of test `testReviewedDateAndCurationLabelsWhenThereIsAReviewedPreprint`

nlisgo (approved):

PR #413, opened 2023-01-10
Add date filter to queue:import

smoqadam:


src/Search/Gearman/Command/ImportCommand.php
```diff
@@ -157,6 +167,12 @@ private function iterateSerializeTask(Iterator $items, string $type, $method = '
                     $items->next();
                     continue;
                 }
+
+                if (!is_null($this->dateFrom) && ($item->getPublishedDate() < $this->dateFrom)) {
```
nlisgo, line 171: We could use `getStatusDate()` if `$item instanceof ArticleVersion` or `$item instanceof ReviewedPreprint` and then we should check if `$item instanceof HasPublishedDate` before using `getPublishedDate()`.

If there is neither `getStatusDate()` or `getPublishedDate` then we should not index that item if the dateFrom option is set. But I don't think we have any content that will have neither.

As this condition and the previous one have the same consequence we could just compound the logic.

src/Search/Gearman/Command/ImportCommand.php
```diff
@@ -65,6 +68,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return;
         }
 
+        if ($input->getOption('dateFrom') !== null) {
+            $this->dateFrom = new \DateTimeImmutable($input->getOption('dateFrom'));
+            if ($this->dateFrom < (new \DateTimeImmutable('-30 days'))) {
+                throw new \Exception('Date must not be bigger than 30 days');
```
nlisgo, line 74: Add use statements for `DateTimeImmutable` and `Exception`. Also, elsewhere we use `DateTimeImmutable::createFromFormat(DATE_ATOM, $input->getOption('dateFrom'))`.

See: https://github.com/elifesciences/search/blob/develop/src/Search/Workflow/ResearchArticleWorkflow.php#L117

nlisgo:
A possible opportunity to test is to introduce a new script such as `bin/ci-import-date-from` and create a check similar to `bin/ci-expected-results` but that checks that only the items indexed are those with the date range. If we did this we would have to allow the script to use a dateFrom which was further in the past than 30 days otherwise we could not consistently get the same number of items indexed with api-dummy.

Let's discuss this tomorrow.

src/Search/Gearman/Command/ImportCommand.php
```diff
@@ -157,6 +172,21 @@ private function iterateSerializeTask(Iterator $items, string $type, $method = '
...
                     continue;
                 }
+
+                if ($item instanceof ArticleVersion || $item instanceof ReviewedPreprint) {
+                    $date = $item->getStatusDate();
+                } elseif ($item instanceof HasPublishedDate) {
+                    $date = $item->getPublishedDate();
+                } else {
+                    $items->next();
+                    continue;
```
nlisgo, line 182: We should only both collecting the date if a dateFrom is set and we should only skip this item if dateForm is set and no date can be determined from the item.

nlisgo (approved):
Small refactor and got rid of the limit on dateFrom as it makes it more testable to not have it in place.

nlisgo:
Agree to let this in as long as we create a followup ticket to introduce the test.

See: https://github.com/elifesciences/search/pull/413#issuecomment-1384555453

PR #414, opened 2023-01-18
Update composer.json to use php7.4

smoqadam:


lsh-0 (approved):
this seems fine

`doctrine/lexer` has had a major version bump that @nlisgo might want to be aware of.

@smoqadam I'll merge this Monday if you don't mind. I'd like to coordinate it with a formula change I'm preparing.

PR #423, opened 2023-03-29
Index public reviews

smoqadam:


src/Search/Api/Elasticsearch/Command/resources/mappings.yaml
```diff
@@ -92,6 +92,12 @@ properties:
     type: text
   authorResponse:
     type: text
+  elifeAssessment:
+    type: text
+  publicReviews:
+    type: text
+  recommendationsForAuthors:
+    type: text
```
nlisgo, line 100: May as well introduce the new fields for now. We can schedule a full re-index at some point with @lsh-0 on hand so that future indexes create the index mapping. Once the mapping is in place it is trivial to target the few VOR PRC for reindex and apply boostings to the new property values.

PR #424, opened 2023-03-29
CI is failing on develop branch

nlisgo:
We can see that for some reason, other than code, the develop branch has begun to fail tests.

@lsh-0 I would appreciate some attention on this, if possible. This was only discovered when @smoqadam began work on https://github.com/elifesciences/search/pull/423

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+CI is failing on develop branch
+
 # eLife Search
 
 This project is using elasticsearch to index content of the eLife articles and provide a full-text search for Journal.


nlisgo:
I'm in the process of deploying search to my local machine with builder so I can attempt to recreate this issue tomorrow and begin attempting to find the source of the problem.

nlisgo:
And, of course, it didn't fail!

So, @lsh-0, you can step down before you step up!

@smoqadam this PR demonstrates that the current build is valid according to tests. Will attempt to progress with this and find out what issues we are introducing.

lsh-0:
No worries. Give me a shout if you need a hand.

PR #427, opened 2023-04-26
Add test to check the most recent reviewed-preprint is the first result

smoqadam:


nlisgo (approved):
There were 2 issues that needed addressing.

- The reviewed-preprints total was broken in api-dummy and was impacted by the `per-page` query parameter. This was fixed by: https://github.com/elifesciences/api-dummy/pull/234
- Because of the above issue and only having one reviewed preprint which was indexed in search the above bug covered up the fact that we weren't offsetting the total of expected items to index with the number of reviewed-preprints which were final. Now in api-dummy we have one reviewed-preprint that is also in the article store so we need to offset the expected total by -1.

This is in much better shape now.

nlisgo:
I've already approved this.

PR #432, opened 2023-06-07
Fix search for terms that has `-`

smoqadam:


nlisgo:
The test is failing even though the fix is in place. This is where commit order can improve confidence in the fix and avoid the risk of a false positive. 

First commit: introduce new api-dummy
Second commit: introduce test which we expect to fail but expresses the bug
Third commit: fix the bug, cross your fingers, tests run green

nlisgo:
I'm going to revert the fix and introduce the api-dummy changes.

PR #439, opened 2023-08-30
Remove gearman

smoqadam:


dev/Dockerfile
```diff
@@ -2,10 +2,9 @@ FROM php:7.4-cli
 
 USER root
 
-RUN apt update && apt install libgearman-dev zip unzip gearman-tools bc procps jq -y && pecl install gearman
+RUN apt update && apt install zip unzip bc procps jq apache2-utils -y
```
nlisgo, line 5: Why do we need to add `apache2-utils`?

src/Search/Console.php
```diff
@@ -293,28 +292,21 @@ public function __construct(Application $console, Kernel $kernel)
         $this->logger = $this->kernel->get('logger');
         $this->root = __DIR__.'/../..';
 
-        if (!defined('GEARMAN_INSTALLED')) {
-            define('GEARMAN_INSTALLED', class_exists('GearmanClient'));
-        }
-
         // Some annotations
-        Register::registerLoader();
+//        Register::registerLoader();
```
nlisgo, line 296: Remove if not needed.

src/Search/Workflow/AbstractWorkflow.php
```diff
@@ -0,0 +1,29 @@
...
+
+    abstract public function index($entity);
+
+    abstract public function insert(string $json, string $id, bool $skipInsert);
+
+    abstract public function postValidate(string $id, bool $skipValidate);
+
+    public function run($entity) {
+        $skipInsert = false;
+        $skipValidate = false;
```
nlisgo, line 17: Are `skipInsert` and `skipValidate` ever true?

src/Search/Workflow/AbstractWorkflow.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace eLife\Search\Workflow;
+
+abstract class AbstractWorkflow
+{
+    abstract public function getSdkClass();
+
+    abstract public function index($entity);
```
nlisgo, line 9: We could make this Model typehint: https://github.com/elifesciences/api-sdk-php/blob/master/src/Model/Model.php

tests/src/Search/Workflow/CollectionWorkflowTest.php
```diff
@@ -5,8 +5,8 @@
 use eLife\ApiSdk\Model\Collection;
 use eLife\Search\Api\ApiValidator;
 use eLife\Search\Api\Elasticsearch\MappedElasticsearchClient;
+use eLife\Search\Workflow\AbstractWorkflow;
```
nlisgo, line 8: Should we just rename the `AbstractWorkflow` `Workflow` as we are removing `eLife\Search\Workflow\Workflow`?

nlisgo (changes requested):
Some comments to a address/discuss.

Also need to talk about the transactional nature of an insert and how to handle the situation if one of the steps doesn't succeed. What if the `postValidate` fails?

dev/Dockerfile
smoqadam: It's needed for the Apache benchmark used in `ci-performance` bash script. With `ab` added to docker we can run CI tests inside the container.

src/Search/Workflow/AbstractWorkflow.php
smoqadam: Yes, they are true in ReviewedPreprintWorkflow

tests/src/Search/Workflow/CollectionWorkflowTest.php
smoqadam: I think it's better to keep it as `AbstractWorkflow`, because it can be recognized even by looking at the filename, and also we have another `Workflow` class in `eLife\Search\Queue`.

smoqadam:
> Some comments to a address/discuss.
> 
> Also need to talk about the transactional nature of an insert and how to handle the situation if one of the steps doesn't succeed. What if the `postValidate` fails?

@nlisgo I've added a retry mechanism if `index`, `insert`, or `postValidate` fail.

nlisgo (approved):

PR #462, opened 2024-02-26
Add test for reindex-reviewed-preprint-dateFrom

smoqadam:


bin/ci-reindex-reviewed-preprints-dateFrom
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -e
+
+#echo "Stopping queue:watch"
```
scottaubrey, line 4: Should this commented code be removed?

scottaubrey (approved):
Overall, I think this is OK, should test the correct thing (as long as api-dummy is correct) and can be merged to gain confidence in the feature.

As a quality/maintenance suggestio: maybe the items should be actually importedinto ES and tested in a way similar to `bin/reimport` where they are imported to a new index?  This could then be inserted anywhere in the list of tests for project_tests.sh, rather than relying more on global things having been run/not run. This could be an improvement for the future, rather than necessary now.

PR #466, opened 2024-04-15
Lazy load articles while adding to queue to prevent memory exhaustion

smoqadam:


src/Search/Gearman/Command/ImportCommand.php
```diff
@@ -165,36 +164,58 @@ public function importBlogArticles()
...
+        $sliceSize = 100;
+        while ($total >= $sliceStart) {
+            foreach ($items->slice($sliceStart, $sliceSize)->toArray() as $item) {
+                // create a Generator to iterate over items in a slice
+                yield $item;
+            }
+            $sliceStart += $sliceSize;
+
+            // This is the magic memory fixer. Without resetting the serialiser, memory continues to grow
+            $denormalizerProperty->setValue($items, $this->getNewSerializer());
```
scottaubrey, line 208: It would be worth fixing this issue inside `Sequence` (or where ever it is in the PHP SDK), even if that's only to reset the denormalizer after each slice call, but perhaps that should be an issue to address later?

scottaubrey (approved):
This matches pretty much my PoC I sent to you, so I'm fairly confident this works. It is just that it's a rather inelegant solution.

As long as there is intention to fix the underlying issue with PHP SDK + denormalizer (via a issue/ticket or whatever), than I'm fine with this as a fix for the immediate needs.

src/Search/Gearman/Command/ImportCommand.php
smoqadam: Thanks Scott, I'll create a ticket for it.

PR #476, opened 2024-08-13
Improve docker

scottaubrey:
These changes primarily are to improve the experience of using docker to run the application, though they also start the process of being able to build a docker container for search.

Improvements:
- Docker-related files are now moved to the top-level directory of the project so we don't need to specify config files when running docker commands.
- A SQS queue watcher service and a gearman worker service are now run in docker compose without dev intervention.
- Included healthchecks for services and proper dependency resolution for services.
- Auto-setup of the queue and KV store indexes.
- The application files are copied into the docker image when built
- Add missing tools to docker image so that tests can be run in docker (rather than currently only on the `search--ci` instance)


scottaubrey:
@ali-tm-amin has tested this out, seems to run OK for him too. 

Most of this is just developer experience stuff: Docker, docker compose. The only real "risk" change in here is to the ci scripts in that they are now more flexible but default to the preexisting values. Jenkins is already running the ci tests in this branch, so that seems to be OK.

So on that basis, I'm going to merge it.

PR #494, opened 2024-10-03
Follow up after remove-gearman to bring clarity to terms and code

scottaubrey:


nlisgo (approved):
I've had a brief look through. I have been able to successfully run project_tests. I have no concerns. I think this is a welcome update. Thanks for your efforts on this. I think we should merge this in ahead of any other major work.

I think to add confidence in this it would be good to deploy to staging and purge the index and see it topped up again before rolling out to prod. What do you think?

scottaubrey:
> I've had a brief look through. I have been able to successfully run project_tests. I have no concerns. I think this is a welcome update. Thanks for your efforts on this. I think we should merge this in ahead of any other major work.
> 
> I think to add confidence in this it would be good to deploy to staging and purge the index and see it topped up again before rolling out to prod. What do you think?

Yeah, that sounds good. Given that prod isn't auto deploying - happy to merge now? @nlisgo

PR #495, opened 2024-10-04
remove weirdness around branches, simplify to just master 

scottaubrey:
elifesciences/issues#8959


scottaubrey:
Note - this targets master branch. We should change the default branch in github first.

